PR #101, opened 2022-02-17
🧹refactor controller

### LINKED ISSUE #50: Use of return: requeue true in the reconciler method
Operator-sdk scaffolded a structure using return requeue, we've removed most of these calls to get the operator working for us. We feel this is not best practice. 

We would like the strucutre to be closer to what operator-sdk originally scaffolded for the memcach example. 

na

**Additional context**
na

jensgrnb:
Closes #50 

Moves logic into submethods and introduces a simple interface to call entrypoint reconcile method on objects.


chris-rock (approved):
Thank you @jensgrnb for addressing the feedback.

I did a minor addition and made most of the methods private since only the `Reconcile` method is really required, everything else should stay internal for now.

Let's make sure we meet the godoc standard and document methods going forward.

PR #106, opened 2022-02-23
WIP: metrics documentation

### LINKED ISSUE #94: Define required metrics to expose it for prometheus
- research best patterns from known operators
- define the metrics that are applicable for mondoo
- add logic to collect the metrics to the operator

harshaisgud:
Enabled servicemonitor for scraping metrics from the operator. Has been tested with prometheus-operator(0.54.0) and kube-prometheus-stack(32.2.1)
fixes #94


config/default/kustomization.yaml
```diff
@@ -22,7 +22,7 @@ bases:
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
 #- ../certmanager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
-#- ../prometheus
+- ../prometheus
```
joelddiaz, line 25: If I'm reading this correctly, un-commenting this effectively tries to turn on metrics monitoring via Promtheus even in environments where Prometheus may not be installed where the ServiceMonitor CRD is not even defined/installed.

Can this instead be more user-controllable where the mondoo-operator only deploys the necessary ServiceMonitor  and ClusterRoleBindings when MondooAuditConfig has asked for `prometheusEnable: True`?

What I have in mind is something similar to how the Webhook support only deploys the cert-manager resources if cert-manager support was enabled in MondooAuditConfig https://github.com/mondoohq/mondoo-operator/pull/135/commits/9fa809044360119241c2a2465b9e76fbcf4194a9 .

harshaisgud: You are correct @joelddiaz , but the reason this PR is on hold is because, the servicemonitor exposes operator metrics like reconcile_total etc, not exactly relavant to Mondoo.
But I agree with your method. Create the service monitor programatically depending on a flag makes perfect sense.

chris-rock:
superseded by https://github.com/mondoohq/mondoo-operator/pull/186

PR #115, opened 2022-02-24
Helm rel fix

harshaisgud:
https://github.com/helm/chart-releaser-action/pull/96 
Currently helm chart release action does not work on tags. There is an open PR addressing this issue. For now , I have switched trigger to pushes on main


chris-rock (approved):
Thank you @harshaisgud

PR #118, opened 2022-02-24
⭐️ updates to allow building with buildah

joelddiaz:
Add new makefile target 'buildah-build' to allow using 'buildah' for
building the container image locally.

As part of that work, buildah gets confused and starts asking questions
about where to fetch the golang:1.17 build container image from. Clarify
that by using a fully qualified name/path for the image
s|golang:1.17|docker.io/library/golang:1.17|

Lastly, add '.vscode' to the .gitignore file for the VSCode users of the
world.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ testbin/*
 
 # editor and IDE paraphernalia
 .idea
+.vscode
 *.swp
 *.swo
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Thank you @joelddiaz

PR #119, opened 2022-02-24
fixes for setting ControllerRef, error/requeue handing, and Status() updates

joelddiaz:
- [x] Move SetControllerReference() to before Create() so the ControllerReference is actually set on the resource(s)
- [x] Improve requeue/error checking when creating ConfigMaps/Deployments/DaemonSets to bubble up any errors and requeue the controller if requested.
- [x] Check if Status() updates are needed via Set() comparison instead of DeepEquals() to avoid potential hotloop if the order of Pods keeps changing between Reconcile() calls.


chris-rock:
Thank you for the improvement @joelddiaz

PR #123, opened 2022-02-25
fixing tags reference in container image build

jensgrnb:


jensgrnb:
Hey @chris-rock, another fixer for the operator-bundle docker build and push step that's failing on main. We believe this is due to the workflow syntax where the tags specify the entire docker URL and not just the semver tag.

chris-rock (approved):
Thank you @jensgrnb

PR #125, opened 2022-02-25
🧹 update helm chart description

harshaisgud:
Updated Version for Change in description


harshaisgud:
Updated chart version to 0.1.3 for update in description of the chart

PR #126, opened 2022-02-28
🧹 restructure olm build and e2e testing

jensgrnb:


.github/workflows/docker-publish.yaml
```diff
@@ -183,37 +183,23 @@ jobs:
...
-  e2e-tests-olm:
-    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
-    needs:
-      - build-operator
-      - build-bundle
-    runs-on: ubuntu-latest
-    steps:
       - name: Start minikube
         uses: medyagh/setup-minikube@master
+
```
chris-rock, line 188: why is the olm test part of the docker publish task and not the e2e workflow?

jensgrnb: To avoid the duplicate operator-sdk installation. Originally I also wanted to reference the same workflow metadata for tags as the build-bundle job, but it's not obvious to me how to reference specific tags from the metadata yet.

chris-rock: I think its confusing, let's extract the olm bundle build and e2e testing into a separate workflow. Therefore the whole `build-bundle` should be moved into its own yaml file. I think it is okay to do the bundle build and e2e test one run, but semantically that is not a container build to me. We only use a container image format as the storage type.

jensgrnb: Okay I'll move them out into their own file

chris-rock: I am not sure I understand. Is it not possible to extract `build-bundle` into a separate workflow?

jensgrnb: It is, but we can't reference "needs xxx" across workflows as far as I understand, and the build-bundle + olm install e2e test needs the operator image to have been pushed successfully.

chris-rock: maybe https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run helps

jensgrnb: Cool, didn't knew that syntax was supported.

jensgrnb:
bundle-build and e2e-test are now in separate workflows. Should be good.

chris-rock (approved):
Great simplification @jensgrnb

PR #128, opened 2022-02-28
🧹 remove minikube references from helm docs

harshaisgud:


harshaisgud:
I cancelled e2e test as there is no change in the code

PR #131, opened 2022-03-01
🐛 fix olm build pipeline

harshaisgud:
Consolidate bundle-push and olm-e2e into Docker workflow. They will be triggered only when `startsWith(github.ref, 'refs/tags/v')`


harshaisgud:
I also uncommented helm-chart releaser action which was disabled by mistake in previous PR while testing helm-release consolidation process.

chris-rock (approved):
Thank you @harshaisgud

PR #135, opened 2022-03-02
Validating webhook support

joelddiaz:
Plumb through pieces to allow standing up a ValidatingWebhookConfiguration via new MondooAuditConfig fields.
- [x] Introduce new 'webhooks' section to MondooAuditConfig
- [x] Add new 'webhook' binary that can be used as an alternative container entrypoint when defining a webhook Deployment
- [x] Add handling to main mondoo-operator to orchestrate bringing up/down the webhook resources

Since the webhooks require TLS certificates, it is not as simple as just setting the webhooks section to `enabled: true`. There is support for using cert-manager to automate requesting the TLS certs (assuming cert-manager is actually installed and available on the cluster). Or you can manually generate TLS certs and place the data in the Secret named mondooNamespace/webhook-server-cert in addition to inserting the CA data into the ValidatingWebhookConfiguration at `webhooks[].clientConfig.caBundle`.


harshaisgud:
@joelddiaz  If I understand this correct we should also be able to use self signed certificates right?

joelddiaz:
> @joelddiaz If I understand this correct we should also be able to use self signed certificates right?

Yes. In fact when using cert-manager, it just asks cert-manager to set up a self-signed CA https://github.com/mondoohq/mondoo-operator/pull/135/files#diff-97606a9f46dabc11e83303db0e9594a06777fe3449c62b09356760a6332535f8R77 . The TLS trust just needs to exist between the contents of ValidatingWebhookConfiguration.webhooks[].clientConfig.caBundle and the TLS certs the webhook Pod mounts via the Secret named webhook-server-cert.

harshaisgud:
Thank You @joelddiaz

controllers/webhooks.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+}
+
+func (n *Webhooks) applyWebhooks(ctx context.Context) (ctrl.Result, error) {
+	r := bytes.NewReader(webhookManifestsyaml)
+	yamlDecoder := yamlutil.NewYAMLOrJSONDecoder(r, 4096)
+	objectDecoder := scheme.Codecs.UniversalDeserializer()
+
+	// Go through each YAML object, convert as needed to Create/Update
+	for {
```
chris-rock, line 238: This is going through the list of objects that we statically created. If we remove resources from webhookManifestsyaml going forward they are not cleaned up. Should we also look for resources with a specific label?

controllers/webhooks.go
```diff
@@ -0,0 +1,418 @@
...
+		} else {
+			webhookLog.Error(err, "Failed to check for existing ValidatingWebhookConfiguration resource")
+			return err
+		}
+	}
+
+	// Doing a DeepEquals comparison of the Webhook is a nightmare. For example: many fields are auto-populated which differ from our
+	// "vanilla" version we generate, and the cert-manager will inject the CA data. The sizes of arrays will be different as well...
+	// So lets just make sure at least the annotation is set and update the object based on that information.
+	// If you want to "force" an update, just change the annotation.
```
chris-rock, line 181: Is the update rolling? I like your approach to just run the update but that mean the system could shutdown a web hook pod for a moment. The service should cover it but just wanted to double-check we have this covered.

controllers/webhooks.go
```diff
@@ -0,0 +1,418 @@
...
+			return ctrl.Result{}, err
+		}
+
+		webhookLog.Info("Decoding object", "GVK", gvk)
+
+		// Cast the runtime to the actual type and delete the resources
+		var genericObject client.Object
+		var conversionOK bool
+		switch gvk.Kind {
+		case "Service":
```
chris-rock, line 385: Those are just the objects we have included in our webhookManifestsyaml right?

controllers/webhooks.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+
+	return nil
+}
+
+func (n *Webhooks) prepareValidatingWebhook(ctx context.Context, vwc *webhooksv1.ValidatingWebhookConfiguration) error {
+
+	switch n.Mondoo.Spec.Webhooks.CertificateConfig.InjectionStyle {
+	case string(mondoov1alpha1.CertManager):
+		if err := n.syncCertManagerIssuer(ctx); err != nil {
```
chris-rock, line 203: should this be covered in a Handler for cert manager? like

```go
type CertManagerHandler struct {}

func (cmh *CertManagerHandler) Setup() error {
….
}
```

Then we could also move the cert manager handling in a separate file. I think it could make the separation between certificates and web hooks a bit more clear

chris-rock (commented):
This is a great addition @joelddiaz 

I added a few smaller questions. Can you please update the development instructions so that cert manager can be setup? Maybe its just a link to https://cert-manager.io/docs/installation/ but it would be nice to name the requirements:

- setup cert-manager via x
- change the AuditConfig value x to y so that we incoming requests are updated

Adding a few kubectl commands to double-check everything is well would help too.

controllers/webhooks.go
joelddiaz: We absolutely could label them and find them that way. Although the OwnerReference should be enough to keep things from being orphaned permanently. I've added a comment in the cleanup codepath that removing resources from webhook-manifest.yaml would need to include custom cleanup code.

controllers/webhooks.go
joelddiaz: Yet it is rolling. So any changes would stand up a second Pod and wait for that to settle before bringing down the previous Pod.

controllers/webhooks.go
joelddiaz: Yes. Just need a way to know what to cast the objects to and to know which functions to call.

controllers/webhooks.go
joelddiaz: I went ahead an added a commit where it separates out the cert manager handling into its own file/struct.

PR #138, opened 2022-03-03
⭐️ adding in service-account reference for daemonset

### LINKED ISSUE #132: DaemonSet is running with default Service Account
**Describe the bug**

The mondoo-client created by the DaemonSet is runnig with the default ServiceAccount of the namespace.

The mondoo-client created by the Deployment is running with its own ServiceAccount `mondoo-operator-workload`.

**To Reproduce**
1. Install mondoo-operator version 0.0.9 (I used the kubectl docs)
2. Wait for DaemonSet to be installed
3. Execute `kubectl get daemonset.apps/mondoo-client -o yaml | grep Account`
This will not return any result.

**Expected behavior**
I would expect the DaemonSet to also use the ServiceAccount `mondoo-operator-workload` or another ServiceAccount, but not the default ServiceAccount.

**Kubernetes:**
 - OS: Ubuntu 20.04
 - Version mondoo-operator 0.0.9
 - Kubernetes version: 1.22.5

jensgrnb:

closes #132


joelddiaz (commented):
I wouldn't expect the user to have to define/configure what the name of the ServiceAccount should be for the DaemonSet that is launched. In what situations would a user select a ServiceAccount for the DaemonSet that is different than the ServiceAccount that is already part of this repository (in config/rbac/workload_service_account.yaml)?

harshaisgud:
@joelddiaz I dont think we should use the workload serviceAccount as it used in a clusterrolebinding that gives access to kubeapi server which the daemonset does not require

joelddiaz:
> @joelddiaz I dont think we should use the workload serviceAccount as it used in a clusterrolebinding that gives access to kubeapi server which the daemonset does not require

In that case, does it make sense to define a new ServiceAccount somewhere under `config/` just for the DaemonSet and attach no Roles/permissions to it (then you can statically assign that ServiceAccount to the DaemonSet and not have to expose anything in the MondooAuditConfig)?

jensgrnb:
Works for me locally

chris-rock (approved):
Thank you @harshaisgud

PR #139, opened 2022-03-03
🔢  add limits on created resources

### LINKED ISSUE #133: all resources managed by mondoo operator are missing required resources (cpu limits and memory)
**Describe the bug**
The mondoo-client Pods created by the Deployment and DaemonSet are running without resources.
This triggers some of the mondoo policies.

**To Reproduce**
1. Install mondoo-operator version 0.0.9
2. Wait for Deployment and DaemonSet to be created
3. Check for resources in spec, e.g.,
`kubectl get deployment.apps/mondoo-client -o yaml | grep -i resources` and
`kubectl get daemonset.apps/mondoo-client -o yaml | grep -i resources`

**Expected behavior**
I would expect both to have resources defined.

As an example:
On our cluster the client consumed ~1.25 CPUs and ~450MB memory.

**Kubernetes**

OS: Ubuntu 20.04
Version mondoo-operator 0.0.9
Kubernetes version: 1.22.5

harshaisgud:
fixes #133


harshaisgud:
minikube in the pipeline only has 2 CPU to allocate to all the pods. Currently finding ways to circumvent this

harshaisgud:
Max supported CPUS for Linux is 2
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

Plans to select runners in roadmap. It is an open issue
https://github.com/github/roadmap/issues/161

Self hosted runners are an option
https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners

.github/workflows/e2e.yaml
```diff
@@ -20,27 +20,31 @@ jobs:
...
+        with:
+          cpus: 4
       - name: Try the cluster !
-        run: kubectl get pods -A
+        run: kubectl describe node minikube
       - name: Build image
         run: |
           export SHELL=/bin/bash
           eval $(minikube -p minikube docker-env)
+          docker pull ghcr.io/mondoohq/mondoo-operator:latest
```
joelddiaz, line 31: Help me understand: what is the purpose of pulling this image if the line right below will do a fresh image build (presumably using the codechanges in the PR under test)?

controllers/nodes.go
```diff
@@ -158,6 +159,12 @@ func (n *Nodes) deamonsetForMondoo(m *v1alpha1.MondooAuditConfig, cmName string)
 						Image:   "mondoolabs/mondoo:latest",
 						Name:    "mondoo-agent",
 						Command: []string{"mondoo", "serve", "--config", "/etc/opt/mondoo/mondoo.yml"},
+						Resources: corev1.ResourceRequirements{
+							Limits: corev1.ResourceList{
+								corev1.ResourceCPU:    resource.MustParse("1.25"),
+								corev1.ResourceMemory: resource.MustParse("500M"),
```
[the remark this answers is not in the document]
harshaisgud: I agree with you. That is what I am planning to do. Set the requests and let the end user decide the limits. Do you have a number in mind for requests? If not I will run Prometheus and check

.github/workflows/e2e.yaml
harshaisgud: Figured it would speed up the build time as most of the initial layers are the same only the code changes. I should be able to optimize this If sharing docker-cache between builds is possible.

joelddiaz: The caching/pre-pull description makes sense to me. I just wanted to make sure it wasn't something that was strictly required (because then that meant that I didn't understand what was happening here...)

harshaisgud: Yeah definitely. Maybe I can create a separate PR for this and not mix these both issues

controllers/nodes.go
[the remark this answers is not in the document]
harshaisgud: I agree with what you are saying, but I am talking more about resource requests for mondoo application itself. Is there a particular number in the documentation(https://mondoo.com/docs/) that I can refer to? If not I will just run the application and see what is the least required resources to scan the kube-system namespace and set that as the requests and let end user decide limits via CRD.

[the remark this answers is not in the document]
joelddiaz: Unfortunately, I don't know those numbers. @chris-rock is this resource usage information gathered/tracked somewhere?

@harshaisgud Setting a baseline requests and allow setting the limits seems like a good initial option. If the MondooAuditConfig doesn't define the limits, would you leave the Pod w/o limits, or place some default limit?

joelddiaz:
@chris-rock and I chatted about this earlier today, and we settled on starting with a default Limit of 1GB / .5 CPU. And setting the Requests to 75% of the limits while still allowing specifying values different from the defaults by updating the MondooAuditConfig.

joelddiaz (commented):
One thing I noticed with the current implementation in the PR is that if I edit the MondooAuditConfig after the node and workload DaemonSet/Deployment have been created, the changes I make to the resource limits/requests are not updated on the already-created DaemonSet/Deployment.

harshaisgud:
@joelddiaz This can be reviewed. I have added a fix for the problem that you have listed

controllers/nodes.go
```diff
@@ -113,6 +114,15 @@ func (n *Nodes) declareDaemonSet(ctx context.Context, clt client.Client, scheme
 	} else if err != nil {
 		log.Error(err, "Failed to get Daemonset")
 		return ctrl.Result{}, err
+	} else if err == nil && !reflect.DeepEqual(found.Spec.Template.Spec.Containers[0].Resources, n.defaultNodeResources()) {
```
joelddiaz, line 117: I think here you want to DeepEqual against n.getNodeResources() instead of n.defaultNodeResources().

By using n.defaultNodeResources(), anytime someone has defined custom resources in the MondooAuditConfig, this codepath will always think an Update() is needed even when it isn't needed because the actual Resources values already match the desired values.

controllers/workloads.go
```diff
@@ -111,6 +113,15 @@ func (n *Workloads) declareDeployment(ctx context.Context, clt client.Client, sc
 	} else if err != nil {
 		log.Error(err, "Failed to get Deployment")
 		return ctrl.Result{}, err
+	} else if err == nil && !reflect.DeepEqual(found.Spec.Template.Spec.Containers[0].Resources, n.defaultNodeResources()) {
```
joelddiaz, line 116: Similar comment here about using n.getWorkloadResources instead of n.defaultNodeResources()

harshaisgud:
Closing this PR and reopening a new one based on main branch as there are many conflicts here.

controllers/nodes.go
chris-rock: DeepEqual is not the correct approach since the data is stored in the struct differently based on how the requirements have been defined. I moved this out in a special function `equalResouceRequirements` and added test cases to make sure we cover this.

controllers/workloads.go
chris-rock: I abstracted and simplified the code. There is no need for two different methods to get default resources requirements.

joelddiaz (commented):
👍 LGTM

PR #144, opened 2022-03-04
⭐️ allow users to overwrite mondoo client image/tag

### LINKED ISSUE #142: Implement system to allow publishing 'stable' container tags in addition to 'latest'
Extend the MondooAuditConfig CRD to allow specifying something besides mondoo:latest ( https://hub.docker.com/r/mondoolabs/mondoo/tags ). In addition to being able to specify tags like 5.28.0 and 5.29.1, allow specifying a moving target like 'stable'.

The default ImagePullPolicy for any container tag not named 'latest' is to be set to 'IfNotPresent'. A node that has once pulled down mondoo:stable would never need to re-pull the container image as long as it remains cached on the node which it is scheduled on. Using a PullPolicy of 'Always' helps a bit, but doesn't solve the situation where the mondoo:stable tag points to a new container image as the Pod will not get restarted if there are no changes to the Pod.Spec.

For the case where a container is using a non-'latest' tag, have the mondoo-operator fetch the SHA of the container image and populate the Deployment/DaemonSet image with the specific container image. This will force a Pod to get rescheduled/updating whenever the container pointed to by 'stable' is updated.

Update the mondoo-operator to ensure that it periodically runs through the full Reconcile() loop (daily?) so that it can fetch the SHA and apply any needed updates to the Deployment/DaemonSets whenever the container image changes.

Something like `return ctrl.Result{Requque: true, RequeueAfter: time.After(time.Now() + time.Day)`

jensgrnb:

closes #142


controllers/mondooauditconfig_controller.go
```diff
@@ -88,14 +90,25 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, err
 	}
 
+	ref, err := name.ParseReference("mondoolabs/mondoo:" + mondoo.Spec.Tag)
```
joelddiaz, line 93: I think hardcoding `mondoolabs/mondoo` in here could complicate some of the ways I've been testing changes to the mondoo-operator repo. I sometimes will build and publish to docker hub (under my own username of course). And having the repo be hard coded would make that kind of testing difficult.

Would it make more sense to expose not just the tag into the MondooAuditConfig.Spec, but the full container image? This would give us options for delivering fixes for someone to test with without having to do a full release. For example, if you were having issues with the node scanning DaemonSet, I could build you `docker.io/myusername/mymondoobuild:latest` with a potential fix and you would be able to quickly test that out and provide feedback by simply updating the MondooAuditConfig resource.

By limiting ourselves to only allowing overriding the tag, we lose some potential flexibility in the future.
I think it may be more helpful to allow specifying the full container image path, and doing the parsing based on that.

@chris-rock WDYT?

controllers/mondooauditconfig_controller.go
```diff
@@ -88,14 +90,25 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+
+	desc, err := remote.Get(ref)
+	imgDigest := desc.Digest.String()
+	repoName := ref.Context().Name()
+	imageUrl := repoName + "@" + imgDigest
+
 	nodes := Nodes{
 		Enable: mondoo.Spec.Nodes.Enable,
 		Mondoo: *mondoo,
+		Image:  imageUrl,
```
joelddiaz, line 106: While this makes 100% sense because today the mondoo-operator only deploys the mondoohq/mondoo container image, I'm thinking about the upcoming Webhook work where the image that is deployed by mondoo-operator is the mondoo-operator itself (with a different entrypoint for the container running the webhook server).

Would it make more sense to have the imageURL/Tag stored for each of MondooAuditConfig.Spec.Nodes and .Spec.Workloads? This would make it cleaner for overriding the Webhook imageURL (whenever that merges), and it would allow narrowing the changes that someone needs to test with if we only want them to test a fix for (for example) node scanning.

chris-rock: I would not tie the Mondoo Client image to nodes or workload since the upcoming scan api will rely on a separate service. I think we could do something like:

```go
spec:
  workloads:
    enable: true
    serviceAccount: mondoo-operator-workload
    image:
      name: docker.io/mondoolabs/mondoo
      tag: 5.29.1
    resources:
      limits:
      requests:
  nodes:
    enable: true
    image:
      name: docker.io/mondoolabs/mondoo
      tag: 5.29.1
    resources:
      limits:
      requests:

```

jensgrnb:
Hey @chris-rock I made the required changes, and tested locally

api/v1alpha1/mondooauditconfig_types.go
```diff
@@ -36,13 +36,20 @@ type MondooAuditConfigData struct {
...
 	Inventory string `json:"inventory,omitempty"`
+	Image     `json:"image,omitempty"`
 }
 
 type Workloads struct {
 	Enable    bool   `json:"enable,omitempty"`
 	Inventory string `json:"inventory,omitempty"`
 	// Replicas               int32  `json:"replicas,omitempty"`
 	ServiceAccount string `json:"serviceAccount,omitempty"`
+	Image          `json:"image,omitempty"`
```
joelddiaz, line 47:
```suggestion
	Image          Image `json:"image,omitempty"`
```

controllers/mondooauditconfig_controller.go
```diff
@@ -88,9 +90,21 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	ref, err := name.ParseReference(image)
+	if err != nil {
+		log.Error(err, "Failed to get container reference")
+	}
+
+	desc, err := remote.Get(ref)
+	imgDigest := desc.Digest.String()
+	repoName := ref.Context().Name()
+	imageUrl := repoName + "@" + imgDigest
+
```
joelddiaz, line 103: Since the MondooAuditConfig is already provided to the Nodes and Workloads structs, can this image fetching logic be moved down into nodes.go and workloads.go?

controllers/mondooauditconfig_controller.go
```diff
@@ -88,9 +90,21 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, err
 	}
 
+	image := mondoo.Spec.Nodes.Image.Name + ":" + mondoo.Spec.Nodes.Image.Tag
+	ref, err := name.ParseReference(image)
+	if err != nil {
+		log.Error(err, "Failed to get container reference")
```
joelddiaz, line 96: Just logging this and allowing the code to proceed leads to a panic when the MondooAuditConfig.Spec.Nodes.Image is not filled out. I believe we want to make specifying the image optional, where if the user leaves the fields blank in the MondooAuditConfig, it uses the default mondoolabs/mondoo:latest image (just like it does today without this PR).
In fact, it probably makes sense to allow setting only the tag (so a user could follow the 'stable' tag of mondoolabs/mondoo), and allowing setting only the 'image' (where 'tag' would be assumed to be 'latest' in that case).

config/samples/k8s_v1alpha1_mondooauditconfig.yaml
```diff
@@ -7,7 +7,13 @@ spec:
   workloads:
     enable: true
     serviceAccount: mondoo-operator-workload
+    image:
+      name: docker.io/mondoolabs/mondoo
+      tag: latest
```
joelddiaz, line 12: Related to my comment where we should allow this to be empty, since 'docker.io/mondoolabs/mondoo' and 'latest' should be considered the defaults, there should be no need to populate this information in the sample mondooauditconfig here.

controllers/mondooauditconfig_controller.go
jensgrnb: Fine with me

controllers/mondooauditconfig_controller.go
jensgrnb: I agree we should return error if this fails, but I don't agree with hiding these fields from the user and hardcoding defaults in, unless this can be done at the struct level.

config/samples/k8s_v1alpha1_mondooauditconfig.yaml
jensgrnb: This goes against what was agreed in previous comment by @chris-rock, where he outlined the structure of the crd.

chris-rock: I agree with @joelddiaz and as discussed the intention was always to have a default defined within the operator and the crd configuration allows users to overwrite it. We do NOT expect users to define the default there.

jensgrnb:
Requested changes made, ready for review.

controllers/nodes.go
```diff
@@ -232,6 +244,33 @@ func (n *Nodes) deamonsetForMondoo(m *v1alpha1.MondooAuditConfig, cmName string)
...
 func (n *Nodes) Reconcile(ctx context.Context, clt client.Client, scheme *runtime.Scheme, req ctrl.Request, inventory string) (ctrl.Result, error) {
+
+	log := ctrllog.FromContext(ctx)
+	var image string
+
+	if n.Mondoo.Spec.Nodes.Image.Name == "" || n.Mondoo.Spec.Nodes.Image.Tag == "" {
+		image = "docker.io/mondoolabs/mondoo:latest"
+	} else {
+		image = n.Mondoo.Spec.Nodes.Image.Name + ":" + n.Mondoo.Spec.Nodes.Image.Tag
+	}
```
joelddiaz, line 255: I think we want to allow mixing and matching image and/or tag overrides. This allows someone to say "I'm okay with the default mondoo container image, but I just want a specific version/tag", and "I have a custom mondoo image that I need to test with, so I want to specify everything".

So something like this:
```suggestion
	mondooImage := "docker.io/mondoolabs/mondoo"
	mondooTag := "latest"
	if n.Mondoo.Spec.Nodes.Image.Name != "" {
		mondooImage = n.Mondoo.Spec.Nodes.Image.Name
	}
	if n.Mondoo.Spec.Nodes.Image.tag != "" {
		mondooTag = n.Mondoo.Spec.Nodes.Image.Tag
	}
	mondooContainer := mondooImage + ":" + mondooTag
```

controllers/workloads.go
```diff
@@ -235,6 +247,33 @@ func (n *Workloads) deploymentForMondoo(m *v1alpha1.MondooAuditConfig, cmName st
 }
 
 func (n *Workloads) Reconcile(ctx context.Context, clt client.Client, scheme *runtime.Scheme, req ctrl.Request, inventory string) (ctrl.Result, error) {
+
+	log := ctrllog.FromContext(ctx)
+	var image string
+
+	if n.Mondoo.Spec.Workloads.Image.Name == "" || n.Mondoo.Spec.Workloads.Image.Tag == "" {
```
joelddiaz, line 254: Same comment here about allowing mixing and matching image and/or tag data from the MondooAuditConfig.

controllers/nodes.go
jensgrnb: Sure. I just don't know anyone who would specify a tag but not an image or visa versa. I'll make the change.

chris-rock: Allowing users to configure either image or tag is great. so that users can just pin the version but do not need to set the image name!

joelddiaz:
@chris-rock just added one commit to this PR to combine the image/tag fetching into a shared utility function. Otherwise the PR was working just fine in my testing.

chris-rock (approved):
Thank you @jensgrnb and @joelddiaz

PR #148, opened 2022-03-07
⭐️ Validating webhook

joelddiaz:
Plumb through pieces to allow standing up a ValidatingWebhookConfiguration via new MondooAuditConfig fields.
- [x] Introduce new 'webhooks' section to MondooAuditConfig
- [x] Add new 'webhook' binary that can be used as an alternative container entrypoint when defining a webhook Deployment
- [x] Add handling to main mondoo-operator to orchestrate bringing up/down the webhook resources

Since the webhooks require TLS certificates, it is not as simple as just setting the webhooks section to `enabled: true`. There is support for using cert-manager to automate requesting the TLS certs (assuming cert-manager is actually installed and available on the cluster). Or you can manually generate TLS certs and place the data in the Secret named mondooNamespace/webhook-server-cert in addition to inserting the CA data into the ValidatingWebhookConfiguration at `webhooks[].clientConfig.caBundle`.


Makefile
```diff
@@ -121,6 +122,8 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 generate-manifests: manifests kustomize ## Generates manifests and pipes into a yaml file
```
chris-rock, line 122: Is the kustomize config missing in deploy, I cannot get this work in developer mode

docs/user-manual-kubectl.md
```diff
@@ -56,6 +56,115 @@ spec:
...
+3. The webhook Deployment should start up, and the ValidatingWebhookConfiguration should be annotated to insert the Certificate Authority data along with the Secret containing the TLS certs will be created by cert-manager (named webhook-serving-cert).3. The webhook Deployment should start up, and the ValidatingWebhookConfiguration should be annotated to insert the Certificate Authority data along with the Secret containing the TLS certs will be created by cert-manager (named webhook-serving-cert).
+
+
+### Manually creating TLS certificates using OpenSSL
+
+1. Create key for your CA.
+``` openssl genrsa -out ca.key 2048 ```
+
+2. Generate certificate for your CA.
+``` openssl req -x509 -new -nodes -key ca.key -sub "/CN=Webhook Issuer" -days 10000 -out ca.crt ```
```
chris-rock: on macOS `-sub` is not supported, I have `LibreSSL 2.8.3` installed

```
openssl req -x509 -new -nodes -key ca.key -subj "/CN=Webhook Issuer" -days 10000 -out ca.crt
```
worked

docs/user-manual-kubectl.md
```diff
@@ -56,6 +56,128 @@ spec:
...
+
+6. Generate the certificate for the webhook server.
+   `openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt -days 10000 -extensions v3_ext -extfile csr.conf`
+
+7. Create the Secret holding the TLS certificate data.
+   `kubectl create secret tls -n mondoo-operator-system webhook-server-cert --cert ./server.crt --key ./server.key`
+
+8. Add the certificate authority as bas64 encoded CA data (`base64 ./ca.crt`) to the `ValidatingWebhookConfiguration` under the `webhooks[].clientConfig.caBundle` field. It should look something like:
+
+```yaml
```
chris-rock, line 162: I am unclear which kubectl command I need to run, do I need to create a new manifest? Do I need to update an existing config?

docs/user-manual-kubectl.md
joelddiaz: Yep, that was a typo. It should be -subj as you noted. I tried these steps with both LibreSSL and OpenSSL and both worked for me.

controllers/mondooauditconfig_controller.go
```diff
@@ -88,6 +96,39 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, err
 	}
 
+	if mondoo.DeletionTimestamp != nil {
+		log.Info("deleting")
```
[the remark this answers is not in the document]
chris-rock: perfect, then we do not need more

docs/user-manual-kubectl.md
chris-rock: Thank you!

docs/user-manual-kubectl.md
chris-rock: Thank you!

chris-rock:
It seems like something is not working with the E2E test https://github.com/mondoohq/mondoo-operator/blob/ba48c8d100c70010bf53f4667d8f852866df2cf8/.github/workflows/e2e.yaml#L36-L43 It takes 60 min and has not even completed.

Makefile
joelddiaz: I think that's just following the default container images defined in:

https://github.com/mondoohq/mondoo-operator/blob/main/config/manager/kustomization.yaml#L13-L16

and the default for the webhook Deployment matches the main mondoo-operator Deployment default (since they use the same container image just with different entrypoints for the container) in:
https://github.com/mondoohq/mondoo-operator/blob/validating-webhook/config/webhook/kustomization.yaml#L11-L12

We can definitely change the default to something more sensible (like :latest), but to test the changes in this PR you will still need to build and publish the mondoo-operator somewhere and set the `IMG` variable when running `make deploy` and `make generate-manifests` to match where the custom build is so that the generated manifests point to the mondoo-operator with the code changes from the PR.

We probably should create a follow-up ticket so that the webhook image can also be modified like what is happening in https://github.com/mondoohq/mondoo-operator/pull/144#discussion_r821477619

joelddiaz:
@chris-rock the DCO checks are finally passing after i force-edited every commit to re-sign each one. I think this is ready to go.

PR #150, opened 2022-03-08
🧹 remove old example

chris-rock:
remove old example that was used initially to build out the operator

Signed-off-by: Christoph Hartmann <chris@lollyrock.com>


joelddiaz (commented):
LGTM 👍

PR #153, opened 2022-03-08
🐛 explicitly remove MondooAuditConfig during teardown

joelddiaz:
'make undeploy' will delete the mondoo-operator Deployment/Pod.

This means the operator will never get a chance to clean up the
finalizer. This means 'make undeploy' hangs waiting for the
namespace to be deleted. And the namespace can't be cleaned up
because the MondooAuditConfig resource is waiting for the finalizer
to be removed.

Add an explicit step to delete the MondooAuditConfig to give the
operator a chance to remove the finalizer, and keep things from
hanging.


joelddiaz:
This should resolve the e2e hangs in #148

PR #154, opened 2022-03-09
🧹 simplify default namespace

### LINKED ISSUE #136: The Operator Namespace should be "mondoo-operator"
The current namespace name of "mondoo-operator-system" is very technically accurate but unusual.  Consider other operator namespace names:

```
openebs                  Active   2y14d
cert-manager             Active   2y6d
klum                     Active   709d
argocd                   Active   687d
mondoo-operator-system   Active   2d7h
```

It would be more natural for the operator namespace to simply be "mondoo"

jensgrnb:

closes #136

### DIFF
--- a/.github/workflows/docker-publish.yaml
+++ b/.github/workflows/docker-publish.yaml
@@ -222,13 +222,13 @@ jobs:
           curl -sSL http://mondoo.io/download.sh | bash
           mv mondoo /usr/local/bin/ 
           echo ${{ secrets.MONDOO_CLIENT }} > creds.yaml
-          kubectl create ns mondoo-operator-system
-          kubectl create secret generic mondoo-client --namespace mondoo-operator-system --from-file=config=creds.yaml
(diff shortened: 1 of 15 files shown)


chris-rock:
this needs to be re-based on the latest main and double-check that the namespace change applies to the web hook work that has been merged

joelddiaz (commented):
This looks good.

I would suggest a rebase on the latest main branch, and updating https://github.com/mondoohq/mondoo-operator/blob/main/config/webhook/kustomization.yaml to also set a default namespace of 'mondoo-operator' there (even though the namespace defined in the files in that directory doesn't really matter since the namespace is overwritten dynamically before being applied, it at least keeps things more consistent throughout the repo).

After that change, running a `make generate-manifests` to build/save the updated controllers/webhook-manifests.yaml file as well.

jensgrnb:
Rebase and make generate-manifests done. This can be merged.

PR #158, opened 2022-03-11
🧹 switch to the new mondoo client image

chris-rock:
We moved the primary container image for Mondoo Client from `mondoolabs/mondoo` to `mondoo/client`. This change uses the new image in the operator.

Signed-off-by: Christoph Hartmann <chris@lollyrock.com>

### DIFF
--- a/controllers/utils.go
+++ b/controllers/utils.go
@@ -23,7 +23,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


joelddiaz (commented):
👍 LGTM

PR #165, opened 2022-03-11
just force-update webhook image:tag

joelddiaz:
Until we come up with a way to keep the operator and the webhook image in sync by default.


chris-rock:
closing this since its part of the release process now, see #156

PR #168, opened 2022-03-11
WIP: 🐛 record mondoo-operator image as env var

joelddiaz:
The mondoo-operator is responsible for launching "itself" when running the webhook Pod. The operator needs a way to know which image it has been built for, so store this an the env var MONDOO_OPERATOR_IMAGE.

Use this env var when getting ready to deploy the webhook to set the container image properly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,8 +115,10 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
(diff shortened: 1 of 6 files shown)


chris-rock:
Instead of using an env var, could we compile the git version into the binary? Then it would just pull the default repo image with the specific tag it was built for?

joelddiaz:
> Instead of using an env var, could we compile the git version into the binary? Then it would just pull the default repo image with the specific tag it was built for?

Maybe I'm missing something here...how do do you translate from having the git sha to the version/tag of the container image?

chris-rock:
We would just set go ldflags. 

We get the version from git:
VERSION = git describe --abbrev=0 --tags
v0.0.13
BUILDTAG git rev-list --count HEAD 
108
 
 And then set the build flags via
 
 ```
-ldflags "-s -w -X go.mondoo.com/mondoo.Version=${VERSION} -X go.mondoo.com/mondoo.Build=${BUILDTAG}"
```

Since version is always the latest released tag, it should just work

joelddiaz:
Closing this in favor of the approach in #190

PR #179, opened 2022-03-14
⭐️ requeue reconcile on weekly basis

### LINKED ISSUE #171: Schedule periodic requeues to allow checking for when the 'latest' tag points to a new container image
When the mondoo-operator has a clean run through Reconcile(), it should requeue itself so that it can periodically check whether the default images or the user overrides (defined in MondooAuditConfig.Spec.Workloads.Image/Tag for example) need updating.

When doing a clean exit from Reconcile() (https://github.com/mondoohq/mondoo-operator/blob/main/controllers/mondooauditconfig_controller.go#L202 ), insert a {Requque: true, RequeueAfter: time.Hour * 24 * 7} to at least let mondoo-operator check weekly for any container image updates.

jensgrnb:

closes #171

### DIFF
--- a/controllers/mondooauditconfig_controller.go
+++ b/controllers/mondooauditconfig_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	_ "embed"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


joelddiaz (commented):
👍 LGTM

PR #180, opened 2022-03-14
⭐️ allow overriding webhooks image

### LINKED ISSUE #164: Allow overriding image/tag for webhook
Similar to the work done in https://github.com/mondoohq/mondoo-operator/pull/144 , add the ability to set an alternative image/tag for the webhook in the MondooAuditConfig CRD.

The main difference from the node/workload is that the default image/tag for the webhook is different than what is used for the node/workload scanning. The webhook uses the mondoo-operator container image (ghcr.io/mondoohq/mondoo-operator:VERSION).

jensgrnb:

closes #164


joelddiaz (commented):
This means I have to update how https://github.com/mondoohq/mondoo-operator/pull/168 works, but ignoring that...
👍 LGTM

chris-rock (approved):
Thank you @jensgrnb

PR #182, opened 2022-03-14
🐛 use input for e2e olm test to allow manual trigger

chris-rock:
uses inputs for olm e2e test to allow manual trigger


joelddiaz (commented):
👍 makes sense to me

PR #189, opened 2022-03-15
⭐️ unit-testing for nodes and workload configuration

### LINKED ISSUE #162: Write test cases for node scanning functions
Similar to how the webhook setup functions have test cases in https://github.com/mondoohq/mondoo-operator/blob/main/controllers/webhooks_test.go , follow that pattern to write up test cases for the code in https://github.com/mondoohq/mondoo-operator/blob/main/controllers/nodes.go

### LINKED ISSUE #163: Write test cases for workload scanning functions
Similar to how the webhook setup functions have test cases in https://github.com/mondoohq/mondoo-operator/blob/main/controllers/webhooks_test.go , follow that pattern to write up test cases for the code in https://github.com/mondoohq/mondoo-operator/blob/main/controllers/workloads.go

jensgrnb:

closes #162, closes #163

### DIFF
--- a/controllers/nodes_test.go
+++ b/controllers/nodes_test.go
@@ -36,7 +36,7 @@ import (
 var _ = Describe("nodes", func() {
 	const (
 		name      = "nodes"
-		namespace = "default"
+		namespace = "nodes-namespace"
 		timeout   = time.Second * 10
 		duration  = time.Second * 10
 		interval  = time.Millisecond * 250
@@ -44,6 +44,16 @@ var _ = Describe("nodes", func() {
 	Context("When deploying the operator with nodes enabled", func() {
 		It("Should create a new Daemonset", func() {
 			ctx := context.Background()
+
+			By("Creating a namespace")
+			ns := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: namespace,
+				},
+			}
+			Expect(k8sClient.Create(ctx, ns)).Should(Succeed())
+
+			By("Creating a secret")
 			secret := &corev1.Secret{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      name,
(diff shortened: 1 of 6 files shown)


controllers/nodes_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+var _ = Describe("nodes", func() {
+	const (
+		name      = "nodes"
+		namespace = "default"
+		timeout   = time.Second * 10
+		duration  = time.Second * 10
+		interval  = time.Millisecond * 250
+	)
+	Context("When deploying the operator with nodes enabled", func() {
```
joelddiaz, line 44: Would be nice to see a test where when there is an existing DaemonSet, and the MondooAuditConfig has node scanning disabled, that the DaemonSet is deleted.

And a test exercising the ability to override the image/tags.

controllers/workloads_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+var _ = Describe("workloads", func() {
+	const (
+		name      = "workloads"
+		namespace = "default"
+		timeout   = time.Second * 10
+		duration  = time.Second * 10
+		interval  = time.Millisecond * 250
+	)
+	Context("When deploying the operator with workloads enabled", func() {
```
joelddiaz, line 44: Similar comment here about ensuring that nothing unexpected is running when Workload.Enabled == false, plus some image/tag test validation.

joelddiaz (commented):
The approach is definitely different than what I'm used to, but it looks sensible to me. How do you go about walking through these tests in a debugger?

controllers/nodes_test.go
jensgrnb: Yep sounds reasonable, I'm still in the discovery phase with ginkgo

joelddiaz:
:+1: LGTM . These first test cases look good enough to merge as-is.

PR #190, opened 2022-03-16
set Version variable during build

joelddiaz:
Take the $VERSION varaible and use it to set the Version variable in controllers/webhooks.go.

This will be used to store the version/tag of the mondoo-operator so that the Deployment/Pod for the webhook is in sync with the mondoo-operator version/tag by default.

You can of course override this by setting MondooAuditConfig.Spec.Webhooks.Image.Tag. And if you were to do a manual 'go build' without setting $VERSION, then it will default to just using the 'latest' tag.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,8 @@ COPY controllers/ controllers/
 COPY pkg/ pkg/
 
 # Build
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Thank you @joelddiaz

PR #191, opened 2022-03-17
Removing SA reference in crd

### LINKED ISSUE #170: Remove serviceAccountName field from MondooAuditConfig.Spec.Workloads
Update the workload deployment functions to not require a user to set the ServiceAccountName field in MondooAuditConfig.Spec.Workloads.

This should work in a similar way to how the .Spec.Nodes section can run without requiring a user to know the name of the ServiceAccount that the Workload Deployment will be using. The mondoo-operator should handle creating the ServiceAccount for the Workload Deployment, and therefor the mondoo-operator will know the name of the ServiceAccount.

jensgrnb:
closes #170 


controllers/mondooauditconfig_controller.go
```diff
@@ -145,7 +145,10 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	workloads := Workloads{
 		Enable: mondoo.Spec.Workloads.Enable,
-		Mondoo: *mondoo,
+		Mondoo: mondoo,
+	}
+	if mondoo.Spec.Workloads.ServiceAccount == "" && mondoo.Namespace == "mondoo-operator" {
```
joelddiaz, line 150: I guess the question I have here is whether or not we support running the mondoo-operator in a namespace besides "mondoo-operator". I took the approach of just reading the namespace that the mondoo-operator is running in dynamically by reading `/var/run/secrets/kubernetes.io/serviceaccount/namespace` (and allowing local development by setting the env var MONDOO_OPERATOR_NAMESPACE). This way if the env var is set, we use that as the "namespace where mondoo-operator is running" or we use the contents of the file.

controllers/mondooauditconfig_controller.go
```diff
@@ -145,7 +145,10 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		mondoo.Spec.Workloads.ServiceAccount = "mondoo-operator-workload"
```
joelddiaz, line 151: "mondoo-operator-workload" should probably just be a constant defined at the top of this file.

joelddiaz (commented):
Just one comment about whether we really want to hardcode the "default" namespace to only ever allow mondoo-operator to run in the `mondoo-operator` namespace.

controllers/mondooauditconfig_controller.go
jensgrnb: Sure, can you point me to an example of how you want this read of an env used?

harshaisgud: https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/ @jensgrnb You can use the downward API to expose pod fields inside container.

jensgrnb: Not sure how this is related, the wish is that the namespace depends on some env?

chris-rock: the service account default is only used if the MondooAuditConfig is placed in `"mondoo-operator` namespace. In any other namespace the user needs to provide a valid service account.

joelddiaz: Using the downward API works just as well. In fact, I'd say that I prefer it over my first approach. This way retrieving the namespace is always done just one way (ie "read the env var").

FWIW, here is what I had implemented previously: https://github.com/mondoohq/mondoo-operator/blob/main/controllers/mondoooperatorconfig_controller.go#L98

jensgrnb: okay I tried using the getNamespace() util function

controllers/mondooauditconfig_controller.go
```diff
@@ -143,7 +144,13 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	workloads := Workloads{
 		Enable: mondoo.Spec.Workloads.Enable,
-		Mondoo: *mondoo,
+		Mondoo: mondoo,
+	}
+
+	ns, _ := getNamespace()
```
joelddiaz, line 150: Check for and handle potential error returned by the function.

joelddiaz (commented):
Any particular reason you are not pursuing the downward API approach to insert the namespace?

jensgrnb:
If you already wrote a subroutine that does what we need, why diverge?

jensgrnb:
This can be reviewed and possibly merged

controllers/mondooauditconfig_controller.go
```diff
@@ -143,7 +144,20 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	}
+
+	namespace, err := getNamespace()
+	if err != nil {
+		log.Error(err, "failed to know which namespace to target")
+		return ctrl.Result{}, err
+	}
+	if err != nil {
+		log.Error(err, "Failed to get namespace")
+	}
```
joelddiaz, line 157: This is redundant as the 'if' clause right above already will catch the error and return out of the function.

jensgrnb:
Suggested changes incorporated, ready for review

controllers/workloads.go
```diff
@@ -260,6 +260,16 @@ func (n *Workloads) Reconcile(ctx context.Context, clt client.Client, scheme *ru
...
+	namespace, err := getNamespace()
+	if err != nil {
+		log.Error(err, "failed to know which namespace to target")
+		return ctrl.Result{}, err
+	}
+
+	if n.Mondoo.Spec.Workloads.ServiceAccount == "" && n.Mondoo.Namespace == namespace {
+		n.Mondoo.Spec.Workloads.ServiceAccount = defaultServiceAccount
+	}
+
```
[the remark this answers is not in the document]
jensgrnb: How would it be ""?

config/samples/k8s_v1alpha1_mondooauditconfig.yaml
```diff
@@ -6,7 +6,6 @@ metadata:
 spec:
   workloads:
     enable: true
-    serviceAccount: mondoo-operator-workload
```
joelddiaz, line 9: Not really related to this line here, but a few lines down where `metrics` is defined should be removed. Controlling metrics is done through the MondooOperatorConfig not MondooAuditConfig.

controllers/workloads.go
```diff
@@ -260,6 +260,18 @@ func (n *Workloads) Reconcile(ctx context.Context, clt client.Client, scheme *ru
...
+		log.Error(err, "failed to determine which namespace mondoo-operator is running in")
+		return ctrl.Result{}, err
+	}
+
+	if n.Mondoo.Spec.Workloads.ServiceAccount == "" && n.Mondoo.Namespace == namespace {
+		n.Mondoo.Spec.Workloads.ServiceAccount = defaultServiceAccount
+	} else {
+		log.Info("MondooAuditConfig in different namespace than mondoo-operator requires setting Workloads.ServiceAccount")
+	}
+
```
joelddiaz, line 274: I believe this should simply be an error/early exit. There is no point in trying to proceed as we don't know which ServiceAccount to bind to the workload Deployment when the MondooAuditConfig is in a different namespace than mondoo-operator.

```suggestion
	if n.Mondoo.Spec.Workloads.ServiceAccount == "" && n.Mondoo.Namespace == namespace {
		n.Mondoo.Spec.Workloads.ServiceAccount = defaultServiceAccount
	}
	if n.Mondoo.Spec.Workloads.ServiceAccount == "" {
	       err := fmt.Errorf("MondooAuditConfig.spec.workloads.serviceAccount cannot be empty when running in a different namespace than mondoo-operator")
		log.Error(err, "ServiceAccount cannot be empty")
		return ctlr.Result{}, err
	}
```

config/samples/k8s_v1alpha1_mondooauditconfig.yaml
jensgrnb: Not sure what is referred here, but I prefer not to change it if it's not related to this PR

joelddiaz: The point I was trying to make is that the config/samples/k8s_v1alpha1_mondooauditconfig.yaml is actually invalid, and I couldn't use it to test out this PR locally.

jensgrnb: Not sure why github isn't marking this as a diff, I branched from main b4 this metrics stuff was removed from this file.

joelddiaz (commented):
👍  LGTM

chris-rock:
I rebased it on latest master since it was difficult to see which change was done on top of main

PR #196, opened 2022-03-18
🧹 remove the full base64 encoded data from docs

joelddiaz:
And just put some boilerplate about where the user needs to copy/paste
the base64 data into.


joelddiaz:
@tas50 I'm totally fine cleaning up the docs so the list of exceptions for the spell checking doesn't need to be so massive.

chris-rock (approved):
Thank you

PR #201, opened 2022-03-21
🐛 clean up generated CRD Description

joelddiaz:
It was picking up unrelated comment lines, so seperate them such that only the lines we want are in the generated Description.


chris-rock (approved):
Thank you @joelddiaz

PR #202, opened 2022-03-21
E2e test of uncommon namespace

### LINKED ISSUE #169: Enable a MondooAuditConfig to exist in a separate namespace than mondoo-operator
Today there are some assumptions made about the MondooAuditConfig existing in the same namespace as where the mondoo-operator runs.

Remove those assumptions and allow the MondooAuditConfig to exist in a different namespace than where the mondoo-operator is running.

This will allow a use-case where UserA can set up mondoo scanning using UserA's mondoo creds in namespace 'user-a'. While UserB can set up additional mondoo scanninng using UserB's mondoo creds in namespace 'user-b'.

To clarify some details based on recent discussions about the need for ServiceAccounts for some of the scanning. It is expected that a default set of ServiceAccounts will exist/be-installed into the default mondoo-operator namespace. This means a user can create an initial MondooAuditConfig in the mondoo-operator namespace without needing to create any ServiceAccounts/Roles/Rolebindings. 

To support creating a MondooAuditConfig resource outside of the default mondoo-operator namespace, there should be docs explaining the process (basically that more than just a MondooAuditConfig is needed for things like workload scanning), with examples of what the ServiceAccounts/Role/Rolebindings would look like. We then need to validate that it is possible to create a MondooAuditConfig outside of the mondoo-operator namespace and ensure that things still work.

jensgrnb:

closes #169


jensgrnb:
This can be reviewed

chris-rock:
Can we rebase this to latest main?

jensgrnb:
Rebased

.github/workflows/e2e.yaml
```diff
@@ -61,6 +61,15 @@ jobs:
           make test/deployment
       - name: Undeploy from minikube
         run: |
+          kubectl delete -f config/samples/k8s_v1alpha1_mondooauditconfig.yaml
```
chris-rock, line 64: That should be a separate step. It is confusing to package that under the undeploy task. I recommend adding a step after undeploy that starts a new deploy and undeploy for the namespace specific test

config/rbac/kustomization.yaml
```diff
@@ -20,8 +20,8 @@ resources:
 #- auth_proxy_role_binding.yaml
 #- auth_proxy_client_clusterrole.yaml
 # Comment the following lines to self mangage roles for the workload
-- workload_service_account.yaml
-- workload_clusterrole.yaml
-- workload_clusterrolebinding.yaml
+- service_account_workloads.yaml
```
chris-rock, line 23: why are we changing this?

chris-rock (changes requested):
Thank you @jensgrnb I added some smaller change requests

.github/workflows/e2e.yaml
```diff
@@ -59,6 +58,18 @@ jobs:
           curl -sSL http://mondoo.io/download.sh | bash
           mv mondoo /usr/local/bin/ 
           make test/deployment
+      - name: Verify Mondoo Operator kustomize manifests in different namespace
+        run: |
+          kubectl delete -f config/samples/k8s_v1alpha1_mondooauditconfig.yaml
+          kubectl create ns not-mondoo-operator
+          kubectl create secret generic mondoo-client --namespace not-mondoo-operator --from-file=config=creds.json
```
joelddiaz, line 65: Add the `rm creds.json` step that was removed above here, or just keep the creds creation/cleanup all together around the `kubectl create secret` command.

.github/workflows/e2e.yaml
```diff
@@ -59,6 +58,18 @@ jobs:
...
+          yq ".metadata.namespace= \"not-mondoo-operator\" | .metadata.name = \"mondoo-operator-nodes\"" config/rbac/service_account_nodes.yaml | kubectl apply -f -
+          yq ".metadata.namespace= \"not-mondoo-operator\" | .metadata.name = \"mondoo-operator-workload\"" config/rbac/service_account_workloads.yaml  | kubectl apply -f -
+          kubectl get clusterrolebinding mondoo-operator-workload -o yaml | yq ".subjects[0].namespace=\"not-mondoo-operator-workload\"" | kubectl apply -f -
+          yq -i ".metadata.namespace= \"not-mondoo-operator\"" config/samples/k8s_v1alpha1_mondooauditconfig.yaml
+          kubectl apply -f config/samples/k8s_v1alpha1_mondooauditconfig.yaml
```
joelddiaz, line 70: You can just do this in a single line
```suggestion
          yq '.metadata.namespace= "not-mondoo-operator"' config/samples/k8s_v1alpha1_mondooauditconfig.yaml | kubectl apply -
```

test/deployment-policy-uncommon-ns.yaml
```diff
@@ -0,0 +1,41 @@
...
+queries:
+  - uid: namespace-exists
+    title: Ensure the namespace exists
+    query: k8s.namespaces.one ( name == "not-mondoo-operator")
+  - uid: crd-created
+    title: Ensure the CRD for MondooScanConfig exitsts
+    query: k8s.apiResources.one( kind == "MondooAuditConfig")
+  - uid: controller-deployed
+    title: Ensure the Controller is deployed
+    query: k8s.deployments.one( name == "mondoo-operator-controller-manager" && namespace == "mondoo-operator" )
```
joelddiaz, line 35: These two queries don't seem to be significant for the non-default-namespace test as this is already covered for the standard namespace test.

joelddiaz:
@chris-rock I added a commit to address the pieces I found during review yesterday. Also while filling in some missing pieces, I found that some changes to the MondooAuditConfig (like the workloads.serviceAccount field) weren't causing an update to the Deployment, so the checking for when the Deployment needs an update has been re-worked.

This should be good for a review now.

.github/workflows/e2e.yaml
chris-rock: To avoid copy paste issues going forward, I would just always doing the complete round. Adding a secret is not an expensive task

.github/workflows/e2e.yaml
```diff
@@ -59,6 +58,18 @@ jobs:
...
+          kubectl get pods -n not-mondoo-operator
+          make test/deployment-uncommon-ns
       - name: Undeploy from minikube
         run: |
           kubectl delete -f config/samples/k8s_v1alpha1_mondooauditconfig.yaml
```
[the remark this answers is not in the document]
chris-rock: Yes lets do this

chris-rock (approved):
@joelddiaz Thank you for adding the changes. This looks great now. Lets update the delete kubectl command and we are ready to go

joelddiaz:
> @joelddiaz Thank you for adding the changes. This looks great now. Lets update the delete kubectl command and we are ready to go

@chris-rock What changes are you saying that should be made to the delete command(s)?

joelddiaz:
> You recommended to make the delete command more specific [#202 (comment)](https://github.com/mondoohq/mondoo-operator/pull/202#discussion_r839996292)

Ah yes, I did implement that in one of my commits https://github.com/mondoohq/mondoo-operator/pull/202/files#diff-a2de7550554c0198ac7f959c78a19ee0996921c98034411de47a7dd49b0c209bR77-R78

W/o the kubectl delete of the MondooAuditConfig the `make undeploy` would hang, so I was forced to add it to get through a complete E2E.

chris-rock:
Ah I was missing the line under the comment. Thank you

PR #203, opened 2022-03-21
0.2.0

joelddiaz:
Synchronize helm chart version with mondoo operator version to 0.2.0 for both.


joelddiaz:
@chris-rock this should be good to go now

chris-rock (approved):
Thank you @joelddiaz

Issue #198, opened 2022-03-18
Allow stubbing/mocking out the image/tag fetching for test cases


joelddiaz:
When running test cases locally, the codepaths will perform the full image/tag-to-SHA256 resolution.

Our test cases shouldn't fail just because DockerHub is down. Modify the way the image resolution is called so that it can be stubbed/mocked out for test cases.

This will keep us from unnecessarily making registry API calls (and eating through any limited quota there), and will allow development of mondoo-operator from a disconnected network environment.

PR #204, opened 2022-03-21
⭐️ Skip resolving mondoo image

### LINKED ISSUE #197: Allow bypassing image/tag resolution
In the event that DockerHub (or whatever container registry is being pulled from) is down. The mondoo-operator will error while setting up Deployments/DaemonSets.
Allow skipping the image/tag fetching via a global mondoo-operator config setting (perhaps using the MondooOperatorConfig CRD).

harshaisgud:
fixes #197,#198 
Skip Mondoo container image resolution when flag is used

### DIFF
--- a/controllers/mondooauditconfig_controller.go
+++ b/controllers/mondooauditconfig_controller.go
@@ -29,7 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"go.mondoo.com/mondoo-operator/api/v1alpha1"
-	mondoov1alpha1 "go.mondoo.com/mondoo-operator/api/v1alpha1"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
@@ -96,17 +96,18 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		log.Error(err, "Failed to get mondoo")
 		return ctrl.Result{}, err
 	}
-	config := &mondoov1alpha1.MondooOperatorConfig{}
-	if err := r.Get(ctx, types.NamespacedName{Name: mondoov1alpha1.MondooOperatorConfigName}, config); err != nil {
+	config := &v1alpha1.MondooOperatorConfig{}
+	if err := r.Get(ctx, types.NamespacedName{Name: v1alpha1.MondooOperatorConfigName}, config); err != nil {
 		if errors.IsNotFound(err) {
 			log.Info("MondooOperatorConfig not found, using defaults")
 		}
-		log.Info(err.Error())
+		log.Error(err, "Failed to MondooOpertorConfig")
+
 	}
 
 	if config.DeletionTimestamp != nil {
-		// Object being deleted; nothing to do
-		return ctrl.Result{}, nil
+		// Going to proceed as if there is no MondooOperatorConfig
+		config = &v1alpha1.MondooOperatorConfig{}
 	}
 
 	if mondoo.DeletionTimestamp != nil {
--- a/controllers/mondooauditconfig_controller.go
+++ b/controllers/mondooauditconfig_controller.go
@@ -101,7 +101,7 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		if errors.IsNotFound(err) {
 			log.Info("MondooOperatorConfig not found, using defaults")
 		} else {
-			log.Error(err, "Failed to MondooOpertorConfig")
+			log.Error(err, "Failed to check for MondooOpertorConfig")
 			return ctrl.Result{}, err
 		}
 	}
@@ -147,7 +147,7 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	nodes := Nodes{
 		Enable:               mondoo.Spec.Nodes.Enable,
-		Mondoo:               *mondoo,
+		Mondoo:               mondoo,
 		MondooOperatorConfig: config,
 	}
 
@@ -161,7 +161,7 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	workloads := Workloads{
 		Enable:               mondoo.Spec.Workloads.Enable,
-		Mondoo:               *mondoo,
+		Mondoo:               mondoo,
 		MondooOperatorConfig: config,
 	}
 
--- a/controllers/mondooauditconfig_controller.go
+++ b/controllers/mondooauditconfig_controller.go
@@ -97,9 +97,9 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, err
 	}
 	config := &mondoov1alpha1.MondooOperatorConfig{}
-	if err := r.Get(ctx, types.NamespacedName{Name: "mondoo-operator-config", Namespace: req.Namespace}, config); err != nil {
+	if err := r.Get(ctx, types.NamespacedName{Name: mondoov1alpha1.MondooOperatorConfigName}, config); err != nil {
 		if errors.IsNotFound(err) {
-			log.Info("MondooOperatorConfig no longer exists")
+			log.Info("MondooOperatorConfig not found, using defaults")
 		}
 		log.Info(err.Error())
 	}
--- a/controllers/mondooauditconfig_controller.go
+++ b/controllers/mondooauditconfig_controller.go
@@ -100,9 +100,10 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	if err := r.Get(ctx, types.NamespacedName{Name: v1alpha1.MondooOperatorConfigName}, config); err != nil {
 		if errors.IsNotFound(err) {
 			log.Info("MondooOperatorConfig not found, using defaults")
+		} else {
+			log.Error(err, "Failed to MondooOpertorConfig")
+			return ctrl.Result{}, err
 		}
-		log.Error(err, "Failed to MondooOpertorConfig")
-
 	}
 
 	if config.DeletionTimestamp != nil {
@@ -121,7 +122,7 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			KubeClient:           r.Client,
 			TargetNamespace:      req.Namespace,
 			Scheme:               r.Scheme,
-			MondooOperatorConfig: *config,
+			MondooOperatorConfig: config,
 		}
 		result, err := webhooks.Reconcile(ctx)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


harshaisgud:
Should this also target mondoo-operator itself? For now now this only skips resolution of the mondoo image @joelddiaz

controllers/nodes.go
```diff
@@ -259,7 +260,8 @@ func (n *Nodes) Reconcile(ctx context.Context, clt client.Client, scheme *runtim
 	log := ctrllog.FromContext(ctx)
 
 	if n.Enable {
-		mondooImage, err := resolveMondooImage(log, n.Mondoo.Spec.Nodes.Image.Name, n.Mondoo.Spec.Nodes.Image.Tag)
+		resolveImage := n.MondooOperatorConfig.Spec.SkipContainerResolution
```
joelddiaz, line 263: The name of the variable sounds reversed to its actual purpose
```suggestion
		skipResolveImage := n.MondooOperatorConfig.Spec.SkipContainerResolution
```

controllers/utils.go
```diff
@@ -38,7 +38,7 @@ const (
 	mondooOperatorTag   = "latest"
 )
 
-func resolveMondooImage(log logr.Logger, userImageName, userImageTag string) (string, error) {
+func resolveMondooImage(log logr.Logger, userImageName, userImageTag string, resolveImage bool) (string, error) {
```
joelddiaz, line 41:
```suggestion
func resolveMondooImage(log logr.Logger, userImageName, userImageTag string, skipResolveImage bool) (string, error) {
```

controllers/utils.go
```diff
@@ -70,15 +68,26 @@ func resolveMondooOperatorImage(log logr.Logger, userImageName, userImageTag str
...
+		imageUrl, err := parseReference(log, mondooContainer)
+		if err != nil {
+			log.Error(err, "Failed to parse reference")
+			return "", err
+		}
+		return imageUrl, nil
+	}
+	return mondooContainer, nil
+}
+
```
joelddiaz, line 90: The naming here should make it clear what the function is doing (returning the image to use for the Pod), and the boolean that controls whether or not to skip the image-to-SHA resolution should indicate its purpose.
 
```suggestion
func getImage(skipResolveImage bool, log logr.Logger, mondooContainer string) (string, error) {
	if !skipResolveImage {
		imageUrl, err := parseReference(log, mondooContainer)
		if err != nil {
			log.Error(err, "Failed to parse reference")
			return "", err
		}
		return imageUrl, nil
	}
	return mondooContainer, nil
}

```

controllers/workloads.go
```diff
@@ -261,7 +262,8 @@ func (n *Workloads) Reconcile(ctx context.Context, clt client.Client, scheme *ru
 	log := ctrllog.FromContext(ctx)
 
 	if n.Enable {
-		mondooImage, err := resolveMondooImage(log, n.Mondoo.Spec.Workloads.Image.Name, n.Mondoo.Spec.Workloads.Image.Tag)
+		resolveImage := n.MondooOperatorConfig.Spec.SkipContainerResolution
```
joelddiaz, line 265:
```suggestion
		skipResolveImage := n.MondooOperatorConfig.Spec.SkipContainerResolution
```

joelddiaz (commented):
While the structs for nodes/webhooks/workloads have been updated to hold the MondooOperatorConfig, nowhere in the mondooauditconfig_controller do I see the MondooOperatorConfig being retrieved and passed into the node/webhook/workloads structs.
This of course means that setting the new skipImageResolution to true doesn't actually cause the node/webhook/workload codepaths to skip the SHA resolution.

controllers/mondooauditconfig_controller.go
```diff
@@ -96,13 +96,11 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, err
 	}
 	config := &mondoov1alpha1.MondooOperatorConfig{}
-	if err := r.Get(ctx, req.NamespacedName, config); err != nil {
+	if err := r.Get(ctx, types.NamespacedName{Name: "mondoo-operator-config", Namespace: req.Namespace}, config); err != nil {
```
joelddiaz, line 99: "mondoo-operator-config" is already a constant defined in https://github.com/mondoohq/mondoo-operator/blob/main/api/v1alpha1/mondoooperatorconfig_types.go#L27 , and there is no need to put a Namespace in your Get() call since MondooOperatorConfig resources are cluster-scoped.

controllers/mondooauditconfig_controller.go
```diff
@@ -96,13 +96,11 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 		if errors.IsNotFound(err) {
 			log.Info("MondooOperatorConfig no longer exists")
-			return ctrl.Result{}, nil
```
joelddiaz: It is okay for the MondooOperatorConfig to not exists, so that log message right above could be worded like "No MondooOperatorConfig found, using defaults."

controllers/mondooauditconfig_controller.go
```diff
@@ -23,15 +23,16 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"go.mondoo.com/mondoo-operator/api/v1alpha1"
+	mondoov1alpha1 "go.mondoo.com/mondoo-operator/api/v1alpha1"
```
joelddiaz, line 32: It looks like this file already was importing this package as `v1alpha`. Just s/mondoov1alpha1/v1alpha1/ in your new lines of code, and you should be able to get rid of importing the same package twice here.
```suggestion
	"go.mondoo.com/mondoo-operator/api/v1alpha1"
```

controllers/mondooauditconfig_controller.go
```diff
@@ -94,6 +95,18 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		log.Error(err, "Failed to get mondoo")
 		return ctrl.Result{}, err
 	}
+	config := &mondoov1alpha1.MondooOperatorConfig{}
+	if err := r.Get(ctx, types.NamespacedName{Name: mondoov1alpha1.MondooOperatorConfigName}, config); err != nil {
+		if errors.IsNotFound(err) {
+			log.Info("MondooOperatorConfig not found, using defaults")
+		}
+		log.Info(err.Error())
```
joelddiaz, line 103: If the error is anything except for IsNotFound() (which is perfectly valid), then we've encountered an unexpected error. We should just log the unexpected error and exit out of Recincile() to try again. 
```suggestion
		if errors.IsNotFound(err) {
			log.Info("MondooOperatorConfig not found, using defaults")
		} else {
		    log.Error(err, "failed to get MondooOperatorConfig)
		    return ctrl.Result{}, err
		 }
```

controllers/mondooauditconfig_controller.go
```diff
@@ -94,6 +95,18 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	}
+
+	if config.DeletionTimestamp != nil {
+		// Object being deleted; nothing to do
+		return ctrl.Result{}, nil
+	}
```
joelddiaz, line 109: This is an interesting case here. There is a MondooOperatorConfig defined in the cluster but it is marked for deletion. We could exit early as you've implemented here, but I think you would want to force a Requeue = true because we still haven't completely processed the MondooAuditConfig, and exiting with `ctrl.Result{}, nil` means the MondooAuditConfig will not be reququed for another pass at Reconcile().

The other alternative, is to zero out the MondooOperatorConfig and let the code run through as if there was no MondooAuditConfig defined. So something like:
```
    if config.DeletionTimestamp != nil {
        // Going to proceed as if there is no MondooOperatorConfig
        config = &mondoov1alpha1.MondooOperatorConfig{}
    }
```

harshaisgud: I like the approach with proceeding instead of requeuing . This ties into #222  to reduce the number of api calls to external repository.

controllers/nodes.go
```diff
@@ -33,10 +33,11 @@ import (
 )
 
 type Nodes struct {
-	Enable  bool
-	Mondoo  *v1alpha1.MondooAuditConfig
-	Updated bool
-	Image   string
+	Enable               bool
+	Mondoo               v1alpha1.MondooAuditConfig
```
joelddiaz, line 37: Why the change from the struct holding `*v1alpha1.MondooAuditConfig` to `v1alpha1.MondooAuditConfig`? In general, for these complex objects like MondooAuditConfig, you just pass around pointers to the objects, and if you truly want a copy of the object, you can perform a `DeepCopy()` of the object (and then pass around the pointer to the copy).

joelddiaz (commented):
This looks good. Just some question about moving away from using a pointer to a MondooAuditConfig towards using a direct embedded MondooAuditConfig.

controllers/utils.go
```diff
@@ -48,18 +48,16 @@ func resolveMondooImage(log logr.Logger, userImageName, userImageTag string) (st
...
 	}
 	mondooContainer := useImage + ":" + useTag
-
-	imageUrl, err := parseReference(log, mondooContainer)
-
+	imageUrl, err := getImage(skipResolveImage, log, mondooContainer)
 	if err != nil {
-		log.Error(err, "Failed to parse reference")
-		return "", err
+		log.Error(err, "Failed to skip resolving image")
```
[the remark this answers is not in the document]
joelddiaz: Whether we skipped or not depends on the skipResolveImage boolean, but what we were really trying to accomplish in this section of code is resolving the image. The error message text should reflect that.

controllers/utils.go
```diff
@@ -70,15 +68,26 @@ func resolveMondooOperatorImage(log logr.Logger, userImageName, userImageTag str
 	}
 	mondooContainer := useImage + ":" + useTag
 
-	imageUrl, err := parseReference(log, mondooContainer)
-
+	imageUrl, err := getImage(skipResolveImage, log, mondooContainer)
 	if err != nil {
-		log.Error(err, "Failed to parse reference")
-		return "", err
+		log.Error(err, "Failed to skip resolving image")
```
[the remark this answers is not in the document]
joelddiaz: Similar comment here about making the error message text explain what we were trying to accomplish when the error was encountered.

controllers/workloads.go
```diff
@@ -34,10 +34,11 @@ import (
 )
 
 type Workloads struct {
-	Enable  bool
-	Mondoo  *v1alpha1.MondooAuditConfig
-	Updated bool
-	Image   string
+	Enable               bool
+	Mondoo               v1alpha1.MondooAuditConfig
```
joelddiaz, line 38: Why the change from pointer to MondooAuditConfig to an embedded struct of MondooAuditConfig here?

controllers/mondooauditconfig_controller.go
```diff
@@ -95,6 +96,20 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		log.Error(err, "Failed to get mondoo")
 		return ctrl.Result{}, err
 	}
+	config := &v1alpha1.MondooOperatorConfig{}
+	if err := r.Get(ctx, types.NamespacedName{Name: v1alpha1.MondooOperatorConfigName}, config); err != nil {
+		if errors.IsNotFound(err) {
+			log.Info("MondooOperatorConfig not found, using defaults")
+		} else {
+			log.Error(err, "Failed to MondooOpertorConfig")
```
joelddiaz, line 104:
```suggestion
			log.Error(err, "Failed to check for MondooOpertorConfig")
```

joelddiaz (commented):
:+1: LGTM

joelddiaz:
@chris-rock this one is good for a second set of eyes to review. It all looks good to me.

chris-rock (approved):
This looks great to me. Thank you @harshaisgud

PR #208, opened 2022-03-21
🐛 create namespace before installing operator bundle

joelddiaz:
The attempt to install mondoo-operator with operator-sdk using the bundle fails because the namespace mondoo-operator doesn't exists.

Move the creation of mondoo-operator Namespace to before the 'operator-sdk run bundle' command.


joelddiaz:
This should resolve these kinds of e2e errors https://github.com/mondoohq/mondoo-operator/runs/5634106495?check_suite_focus=true

chris-rock (approved):
Thank you @joelddiaz

PR #210, opened 2022-03-22
🧹 fixes doc config and command references

scottford-io:
This PR fixes a couple of issues I ran into while following the docs...

## error validating data: apiVersion not set

```bash
kubectl apply -f mondoo-config.yaml
error: error validating "mondoo-config.yaml": error validating data: apiVersion not set; if you choose to ignore these errors, turn validation off with --validate=false
```
Missing `apiVersion: admissionregistration.k8s.io/v1`. 

## No resources found in default namespace.

```bash
kubectl get mondooauditconfigs
No resources found in default namespace.
```

Missing `-A`

```bash
kubectl get mondooauditconfigs -A
NAME            AGE
mondoo-client   20m
```

Signed-off-by: Scott Ford <scott@scottford.io>

### DIFF
--- a/docs/user-manual-helm.md
+++ b/docs/user-manual-helm.md
@@ -73,7 +73,7 @@ mondooauditconfigs.k8s.mondoo.com   2022-01-14T14:07:28Z
 Then make sure a configuration for the Mondoo Client is deployed:
 
 ```bash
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Thank you @scottford-io

PR #212, opened 2022-03-22
⭐️ CRD status update

### LINKED ISSUE #141: Report overall status into the MondooAuditConfig CR
Start with extending the MondooAuditConfigStatus to hold fields where the mondoo-operator can report status based on the health of the deployed DaemonSets/Deployments.

A first implementation should at least be able to check that the Deployment and DaemonSet are Ready/Running. Something like the golang equivalent of:

```
jdiaz@joels-air mondoo-operator % kubectl get deployment -n mondoo-operator-system mondoo-client -o json | jq -r '.status.conditions[] | select(.type=="Available") | .status'        
True
```

And for the DaemonSet `kubectl get daemonset -n mondoo-operator-system mondoo-client -o json | jq -r .status.desiredNumberScheduled` should equal `.status.numberReady`.
 
And if all the Deployments and DaemonSets are healthy, then bubble that up as a high-level MondooAuditConfig.Status when all items are healthy. If a single one is unhealthy, then the entire status should reflect the unhealthy state.

```
// MondooAuditConfigStatus defines the observed state of MondooAuditConfig
type MondooAuditConfigStatus struct {
	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
	// Important: Run "make" to regenerate code after modifying this file

	// Nodes store the name of the pods which are running mondoo instances
	Pods []string `json:"pods,omitempty"`

        OverallStatus string `json:"overallStatus,omitempty"`
}
```
The `OverallStatus` can be limited to a the options of Healthy or Degraded.

Perform fine-grained report for the individual items being monitored by using Conditions in the MondooAuditConfig.Status similar to what you find in a Deployment:

```
jdiaz@joels-air mondoo-operator (validating-webhook *) % kubectl get deployment -n mondoo-operator-system mondoo-client -o json | jq -r .status.conditions
[
  {
    "lastTransitionTime": "2022-03-04T17:18:34Z",
    "lastUpdateTime": "2022-03-04T17:18:34Z",
    "message": "Deployment has minimum availability.",
    "reason": "MinimumReplicasAvailable",
    "status": "True",
    "type": "Available"
  },
  {
    "lastTransitionTime": "2022-03-04T17:18:32Z",
    "lastUpdateTime": "2022-03-04T17:18:34Z",
    "message": "ReplicaSet \"mondoo-client-5b6cff55f9\" has successfully progressed.",
    "reason": "NewReplicaSetAvailable",
    "status": "True",
    "type": "Progressing"
  }
]
```

This would be something like type=NodeScanningAvailable with a status of True/False based on that individual component health.
And a type=APIScanningAvailable for the API scanning Deployment, etc. More specifically follow the k8s format with a struct to hold the condition results that looks like:
```
type MondooAuditConfigCondition struct {
	// Type is the specific type of the condition
	// +kubebuilder:validation:Required
	// +required
	Type AudicConfigConditionType `json:"type"`
	// Status is the status of the condition
	// +kubebuilder:validation:Required
	// +required
	Status corev1.ConditionStatus `json:"status"`
	// LastProbeTime is the last time we probed the condition
	LastProbeTime metav1.Time `json:"lastProbeTime,omitempty"`
	// LastTransitionTime is the last time the condition transitioned from one status to another.
	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
	// Reason is a unique, one-word, CamelCase reason for the condition's last transition
	Reason string `json:"reason,omitempty"`
	// Message is a human-readable message indicating details about the last transition
	Message string `json:"message,omitempty"`
}
```

An example where the node scanning is healthy, and the API scanning is unhealthy would look like:
```
jdiaz@joels-air mondoo-operator % kubectl get mondooauditconfig -n mondoo-operator-system mondoo-client -o json | jq -r .status.conditions
[
  {
    "lastTransitionTime": "2022-03-04T17:18:34Z",
    "lastUpdateTime": "2022-03-04T17:18:34Z",
    "message": "Node scanning deployed successfully.",
    "reason": "DaemonSetHealthy",
    "status": "True",
    "type": "NodeScanningAvailable"
  },
  {
    "lastTransitionTime": "2022-03-04T17:18:32Z",
    "lastUpdateTime": "2022-03-04T17:18:34Z",
    "message": "API scanning deployment degraded.",
    "reason": "Desired replicas not equal to running replicas.",
    "status": "False",
    "type": "APIScanningAvailable"
  }
]
```

Some concrete examples of the kinds of things we would want a condition for:
1. If the user tries to enable metrics reporting, but the Prometheus CRDs are not installed on the cluster

jensgrnb:

closes #141


jensgrnb:
Hey @joelddiaz can you take a look at here, and tell us if this approach is similar to what you had in mind in the issue #141?

controllers/nodes.go
```diff
@@ -183,6 +183,18 @@ func (n *Nodes) deamonsetForMondoo(m *v1alpha1.MondooAuditConfig, cmName string)
 						Name:      "mondoo-agent",
 						Command:   []string{"mondoo", "serve", "--config", "/etc/opt/mondoo/mondoo.yml"},
 						Resources: getResourcesRequirements(m.Spec.Nodes.Resources),
+						StartupProbe: &corev1.Probe{
```
joelddiaz, line 186: Not sure if the StartupProbe is strictly necessary (is the mondoo client known to be slow to start up?). But I believe what we want is an ongoing ReadinessProbe.

IIUC, the `mondoo status` command will fail if the credentials are determined to be useless? So we would want a ReadinessProbe that periodically runs these checks (perhaps every 30 minutes or 1 hour to not make too many Mondoo API calls).

joelddiaz (commented):
This is looking good. The missing piece here is implementing the MondooAuditConfig.Status.Conditions[] list to allow reporting on the health of the individual node/workload/webhook pieces that are enabled in the MondooAuditConfig resource.

Then the OverallStatus can be Healthy if all of node/workload/webhook pieces are healthy. And an OverallStatus of Degraded will indicate for someone to look into the MondooAuditConfig.Status.Conditions[] for more details on which component is degraded.

api/v1alpha1/mondooauditconfig_types.go
```diff
@@ -86,8 +87,49 @@ type MondooAuditConfigStatus struct {
...
+	// NodeScanningAvailable is used to indicate  NodeScanning is available.
+	NodeScanningAvailable MondooAuditConfigConditionType = "NodeScanningAvailable"
+	// APIScanningAvailable is used to indicate  ApiScanning is available .
+	APIScanningAvailable MondooAuditConfigConditionType = "APIScanningAvailable"
+	// NodeScanningUnAvailable is used to indicate  NodeScanning is Unavailable.
+	NodeScanningUnavailable MondooAuditConfigConditionType = "NodeScanningUnavailable"
+	// APIScanningAvailable is used to indicate  ApiScanning is Unavailable.
+	APIScanningUnavailable MondooAuditConfigConditionType = "APIScanningUnvailable"
+)
+
```
harshaisgud, line 132: These Conditions are not required if we are adding the deployment and the daemonset conditions to the overall status.

jensgrnb:
Hey @joelddiaz we are still having trouble with the CI and the probes, but the jest of the PR is here to be reviewed. We've tested locally and it seems to be working.

joelddiaz (commented):
I would suggest looking at https://github.com/mondoohq/mondoo-operator/blob/main/controllers/service_monitor.go#L124-L155 for an example of setting/clearing conditions. What we want here is individual ConditionTypes to capture the individual health of node/webhook/workload. And then once all those Conditions have been calculated, we can know what the OverallStatus should be.

controllers/workloads.go
```diff
@@ -183,6 +184,18 @@ func (n *Workloads) deploymentForMondoo(m *v1alpha1.MondooAuditConfig, cmName st
...
 						Command:   []string{"mondoo", "serve", "--config", "/etc/opt/mondoo/mondoo.yml"},
 						Resources: getResourcesRequirements(m.Spec.Workloads.Resources),
+						ReadinessProbe: &corev1.Probe{
+							ProbeHandler: corev1.ProbeHandler{
+								Exec: &corev1.ExecAction{
+									Command: []string{"mondoo", "status"},
+								},
+							},
+							InitialDelaySeconds: 1,
+							PeriodSeconds:       1800,
```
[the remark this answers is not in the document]
harshaisgud: https://mondoo.com/docs/references/cli/mondoo_status/

api/v1alpha1/mondooauditconfig_types.go
```diff
@@ -86,8 +87,49 @@ type MondooAuditConfigStatus struct {
 
 	// Pods store the name of the pods which are running mondoo instances
 	Pods []string `json:"pods,omitempty"`
+	// OverallStatus store the status of the objects running mondoo instances
+	OverallStatus string `json:"overallstatus,omitempty"`
+	// DaemonsetConditions store the conditions for the mondoo daemonset
+	DaemonsetConditions []appsv1.DaemonSetCondition `json:"daemonsetconditions,omitempty"`
+	// DeploymentConditions store the conditions for the mondoo deployment
+	DeploymentConditions []appsv1.DeploymentCondition `json:"deploymentconditions,omitempty"`
```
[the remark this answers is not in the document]
jensgrnb: There's a lot of available fields from the base objects, could you maybe commit an example of what you have in mind how to calculate for instance the apiscanningdegraded type.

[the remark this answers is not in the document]
joelddiaz: You can start with a simple approach for the initial implementation
```
Get(ctx, types.NamesapacedName{...}, currentDeployment)
if currentDeployment.Status.ReadyReplicas < currentDeployment.Spec.Replicas {
  /* then set the condition */
} else {
/* clear the condition */
}
```
You do that check for each of the Deployments/DaemonSets for the node/webhook/workload scanning.

If in the future we need more complex probing of the Deployment/DaemonSet, that can be extended later.

controllers/nodes.go
```diff
@@ -154,6 +155,11 @@ func (n *Nodes) declareDaemonSet(ctx context.Context, clt client.Client, scheme
 		}
 		return ctrl.Result{Requeue: true}, err
 	}
+	config := n.Mondoo.DeepCopy()
+	updateNodeConditions(config, found)
+	if err := UpdateMondooAuditStatus(ctx, clt, &n.Mondoo, config, log); err != nil {
+		return ctrl.Result{}, err
+	}
```
joelddiaz, line 162: Rather than have each of nodes.go and workloads.go be responsible for calling Status().Update(), do a DeepCopy() of the MondooAuditConfig near the top of the mondooauditconfig_controller.go to save it as `originalMondooAuditConfig`, then you can pass in the MondooAuditConfig into the node/webhook/workload structs as it presently being done, and allow each of them to add/update their specific Condition they are responsible for reporting on.

Then near the end of mondooauditconfig_controller.go, you can do a single  call to 
```
UpdateMondooAuditStatus(ctx, client, originalMondooAuditConfig, mondoo)
```

This way as each of nodes.Reconcile(), webhooks.Reconcile(), and workloads.Reconcile() is called, they each just modify the mondoo.Status.Conditions, and you do a single Update() call before exiting the mondooauditconfig_controller.go.Reconcile()

controllers/nodes.go
```diff
@@ -183,6 +190,18 @@ func (n *Nodes) deamonsetForMondoo(m *v1alpha1.MondooAuditConfig, cmName string)
...
+							ProbeHandler: corev1.ProbeHandler{
+								Exec: &corev1.ExecAction{
+									Command: []string{"mondoo", "status"},
+								},
+							},
+							InitialDelaySeconds: 1,
+							PeriodSeconds:       1800,
+							TimeoutSeconds:      10,
+							SuccessThreshold:    1,
+							FailureThreshold:    1,
```
joelddiaz, line 203: I feel like my comments about this from before got lost.

We want to allow transient failures to not immediately mark the Pod as NotReady. So a FailureThreshold of `2` will be more forgiving on the off-chance that the Mondoo API was inaccessible the exact moment the ReadinessProbe was run.

controllers/nodes.go
jensgrnb: Hey Joel, not sure if this is the right way to go, the reconcile methods on the nodes, workloads etc. will have to return these configs and then in the end we'll have to merge them all together to do a single update call. Seems overly complicated compared to the current solution.

joelddiaz: @jensgrnb sorry if my explanation wasn't clear. What I mean is that at the top of the mondooauditconfig Reconcile() you DeepCopy() the MondooAuditConfig to save what the "original" object looked like.

Then inside of nodes/webhook/workloads Reconcile() they can add/update the .Status.Conditions[] as needed. And since nodes/webhook/workloads is working on a reference to the MondooAudicConfig object, each of the nodes/webhook/workloads changes are reflected in the mondooauditconfig_controller's copy of the MondooAuditConfig.

Then in the mondooauditconfig Reconcile() you do a single check to see if the Status needs to be updated (and update it if necessary).

If each nodes/webhook/workloads does their own update to Status, then what can happen is this:
1) nodes does the DeepCopy() of the original MondooAuditConfig and does a Status().Update() which creates a new/updated version of the MondooAuditConfig in etcd.
2) workloads has a stale copy of the MondooAuditConfig (which is missing the changes that nodes made to the conditions), and if/when workloads tries to DeepCopy(), then Status().Update() for itself, then either you get an error (because workloads has an out-of-date version of the MondooAuditConfig) or the Update() succeeds and workloads wipes away the changes that nodes made to the Status (I think you actually get an error for trying to Update() a stale copy).
3) same sequence can happen for webhooks trying to Status().Update().

jensgrnb: Latest commit to this PR has a global update call, please check if this is what you expected. 

I don't think we've observed the concerns you've listed here, but there are upsides and downsides to both approaches, I just think the status update should happen as close as possible to the actual update of the objects -- in the x.declareDeployments/Daemonsets whatever. If they don't have an updated version of the crd at that point we should do a get call. 

It's also a more scalable solution than to accumulate everything at the end of the top-level reconcile method.

jensgrnb:
Works for me locally, ready for review

controllers/workloads.go
```diff
@@ -184,16 +186,6 @@ func (n *Workloads) deploymentForMondoo(m *v1alpha1.MondooAuditConfig, cmName st
...
-						ReadinessProbe: &corev1.Probe{
-							ProbeHandler: corev1.ProbeHandler{
-								Exec: &corev1.ExecAction{
-									Command: []string{"mondoo", "status", "--config", "/etc/opt/mondoo/mondoo.yml"},
-								},
-							},
-							InitialDelaySeconds: 10,
-							PeriodSeconds:       300,
-							TimeoutSeconds:      5,
-						},
```
joelddiaz, line 196: Keep what is in main.

controllers/nodes.go
joelddiaz: To address your concern about missing the chance to do the Status().Update(), we can put a `defer` after making the DeepCopy() to save the original MondooAuditConfig, and use named return variables (reconcileResult and reconcileError) for the mondooauditconfig_controller Reconcile().

Then in the defer, do the check to see if the Status().Update() is needed, and if the update fails, and the named `reconcileError` == nil, set reconcileError to the error received while doing the Update(). But leave reconcileError alone if it isn't nil because Status Updates aren't nearly as important as the core reconciliation errors (but still log the fact that we failed to update Status).

jensgrnb: Maybe, but to me, this is a bad sign when you have to write all this handholding syntax. It bloats the codebase, distracts other people when reading the codeflow and makes the code harder to expand.

controllers/workloads.go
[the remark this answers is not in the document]
chris-rock: This was solved separately in #230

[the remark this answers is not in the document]
chris-rock: This was solved separately in #230

controllers/mondooauditconfig_controller.go
```diff
@@ -129,6 +129,7 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 	}
 
+	config := mondoo.DeepCopy()
```
chris-rock, line 132: to avoid confusion, I think we should rename `mondoo` to `mondooAuditConfig` and `config` to `mondooAuditConfigClone` this would make it more clear from reading through the code

chris-rock: when we do the clone here, why are we passing in the original config to the structs?

jensgrnb: mondoo --> mondooAuditConfig done
config --> mondooAuditConfigCopy done

The orignal has always been passed to the structs, we didn't have a copy before.

joelddiaz (commented):
:+1: LGTM

PR #216, opened 2022-03-22
🧹 remove Webhook Service/Deployment from generated manifests

joelddiaz:
The Service and Deployment resources used for setting up a ValidatingWebhookConfiguration are very much static except for the Namespace they get created into. Move those resources out of the kustomize pipeline and just integrate them into the webhook handling code like what is done for node/workload scanning resources.

This just involved manually translating the previous webhook Service and Deployment yaml definitions into golang struct/types.

Change the name of the webhook service to match the name of the MondooAuditConfig. This should allow multiple MondooAuditConfig resources sharing a namespace. Update the cert-manager code to set up the DNS names to match the new Service naming scheme.


chris-rock (approved):
Thank you @joelddiaz

joelddiaz:
@chris-rock I rebased this PR to pull in the changes from #218 , so the E2E stuff now runs to completion.

PR #218, opened 2022-03-23
🐛 fake some event data when running via act

joelddiaz:
It appears that act doesn't set up enough of a fake environment which
can cause some github action workflows that work on the real github to
fail when run through act.

Create some fake event data so that the checks against specific comments
in the spell-check workflow can succeed when run locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -250,8 +250,14 @@ test/fmt:
 
 .PHONY: test/github-actions
 test/github-actions:
-	act -n --container-architecture linux/amd64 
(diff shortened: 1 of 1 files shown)


joelddiaz:
These changes resolved the issues I reproduced locally. But they can also be seen when a github action actually calls into our Makefile to run through 'act'. https://github.com/mondoohq/mondoo-operator/runs/5662258258?check_suite_focus=true 

The "real" github actions run just fine, as can be seen in #216 
image

chris-rock (approved):
Thank you @joelddiaz

PR #222, opened 2022-03-24
Resolving Image at MondooAuditConfig Level

### LINKED ISSUE #199: Audit mondoo-operator for unnecessary early exits
There are some places in the codepaths where an exit is made with a Requeque = true set.
In many cases this isn't needed and the Reconcile() can continue normally.

For example in a sequence like https://github.com/mondoohq/mondoo-operator/blob/main/controllers/nodes.go#L64-L70 , it is okay to continue forward and create/update the DaemonSet if needed. The DaemonSet/Pod should be unable to run if any volume that it needs to mount in isn't defined. Once the ConfigMap/volume appears, the DaemonSet/Pod should launch successfully.

While in general the Requeue set to true is harmless, the Reconcile() loops do make API calls to remote registries to resolve the SHA256, and the early exits from the Reconcile() cause more registry API calls than is strictly necessary.

harshaisgud:
fixes #199

### DIFF
--- a/controllers/mondooauditconfig_controller.go
+++ b/controllers/mondooauditconfig_controller.go
@@ -41,7 +41,8 @@ const (
 // MondooAuditConfigReconciler reconciles a MondooAuditConfig object
 type MondooAuditConfigReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme      *runtime.Scheme
+	ImageDigest string
 }
 
 // Embed the Default Inventory for Daemonset and Deployment Configurations
@@ -127,10 +128,18 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			return ctrl.Result{}, err
 		}
 	}
+	if r.ImageDigest == "" {
+		mondooImage, err := resolveMondooImage(log, mondoo.Spec.Nodes.Image.Name, mondoo.Spec.Nodes.Image.Tag)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		r.ImageDigest = mondooImage
+	}
 
 	nodes := Nodes{
 		Enable: mondoo.Spec.Nodes.Enable,
 		Mondoo: *mondoo,
+		Image:  r.ImageDigest,
 	}
 
 	result, err := nodes.Reconcile(ctx, r.Client, r.Scheme, req, string(dsInventoryyaml))
@@ -144,6 +153,7 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	workloads := Workloads{
 		Enable: mondoo.Spec.Workloads.Enable,
 		Mondoo: *mondoo,
+		Image:  r.ImageDigest,
 	}
 
 	result, err = workloads.Reconcile(ctx, r.Client, r.Scheme, req, string(deployInventoryyaml))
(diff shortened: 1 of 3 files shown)


harshaisgud:
@joelddiaz This change moves image resolution to mondooauditconfig controller, thereby reducing the number of API calls to external image repo

chris-rock:
I'll try to close and re-open the pull request to see if Github restarts the tests

controllers/mondooauditconfig_controller.go
```diff
@@ -144,6 +153,7 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	workloads := Workloads{
 		Enable: mondoo.Spec.Workloads.Enable,
 		Mondoo: *mondoo,
+		Image:  r.ImageDigest,
```
joelddiaz, line 156: Here the ImageDigest being assigned is the one fetched from the contents of Spec.Nodes.Image.* (not Spec.Workloads.Image.*). You can't assume that they are set to the same image/tag set.

controllers/mondooauditconfig_controller.go
```diff
@@ -127,10 +128,18 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			return ctrl.Result{}, err
 		}
 	}
+	if r.ImageDigest == "" {
+		mondooImage, err := resolveMondooImage(log, mondoo.Spec.Nodes.Image.Name, mondoo.Spec.Nodes.Image.Tag)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		r.ImageDigest = mondooImage
```
joelddiaz, line 136: Storing this value in the struct MondooAuditConfigReconciler means that if there are two MondooAuditConfig resources in the cluster, and they are being Reconciled in parallel (which can/does happen), then the first MondooAuditConfig that assigns r.ImageDigest will also affect the second MondooAuditConfig unless/until r.ImageDigest is returned to the empty string at the bottom of this function.
This is a racy implementation that will have subtle and hard-to-debug behaviors.

joelddiaz (commented):
The idea with #199 is to eliminate the unnecessary places in the codepath where it does a `return ctrl.Result{Requeue: true}, nil`.

There are really two options to try to address the excessive external API calls to the container registry.
1) Remove the points in the codepath that perform the returns with Requeue set to true (and I don't mean simply not set Requeue to true, I mean remove the `return` entirely and allow the code to continue running).

If we leave the early `return` code, then the alternative is:
2) Implement some new long-lived structures to cache the image/tag -> SHA fetching and only try to refresh the image/tag/sha if the image/tag/sha entry was fetched more than X minutes ago.

controllers/mondooauditconfig_controller.go
harshaisgud: @joelddiaz Implementing a long lived struct is going to be the same as storing it `MondooAuditConfigReconciler` though unless we identify each custom resource with a uuid.

joelddiaz:
Another approach (as suggested by @chris-rock ) was to store the SHAs in the MondooAuditConfig.Status along with a timestamp recording the last time the SHAs were fetched. This would allow the Reconcile() codepaths to see if the SHAs were recently fetched, then skip re-fetching them until the lastImageResolveTimestamp happened more than (for example) 1 week ago.

So perhaps something like:
```
Status:
  ImageSHAs:
  - containerImage: docker.io/mondoo/mondoo-client:latest
    sha: docker.io/mondoo/mondoo-client@sha256:asfdasfdasfdfasdf
    lastImageResolveTimestamp: 2022-03-30-22:00...
  - containerImage: ghcr.com/mondoohq/mondoo-operator:latest
    sha: ghcr.com/mondoohq/mondoo-operator@sha256:asfdasfdasdfasfd
    lastImageResolveTimestamp: 2022-03-29-09:00...
```

And then the `getImage()` function can reference this list of cached SHAs and reuse them if they are recent, and actually perform the full fetch and updating of the Status to update the cache for future Reconcile()s.

PR #230, opened 2022-03-29
⭐️ Readiness Probe

### LINKED ISSUE #63: Add health checks to Mondoo Client pods
Current policy only checks for existence of deployment and daemonset. Needs to be extended to check if the containers have started up correctly.

Use Mondoo to scan container runtime logs files for correct startup messages.

Just parse files using awk  and compare . It is cumbersome . Better to include as part of policy.

**Additional context**
Mondoo currently does not allow parsing journalctl files

harshaisgud:
fixes #63

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -3,7 +3,7 @@ on:
   pull_request:
   push:
     paths-ignore:
-      - 'docs/**'
+      - "docs/**"
     branches:
       - "main"
     tags: ["v*.*.*"]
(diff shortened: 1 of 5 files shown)


joelddiaz:
@harshaisgud well that took a while, but I believe I see what is happening. The readiness probe runs `mondoo status`, and if you read the output of `mondoo status --help` you see:
```
      --config string      config file (default is $HOME/.config/mondoo/mondoo.yml)
```
In the mondoo client container, $HOME is `/root`, and the actual `mondoo.yml` file lives in `/etc/opt/mondoo/mondoo.yml`. So in the same way that the main process that gets executed in the Deployment as `mondoo serve --config /etc/opt/mondoo/mondoo.yml`, the readiness probe should also specify `--config /etc/opt/mondoo/mondoo.yml`.

Or the alternative is to mount the file into `/root/.config/mondoo/mondoo.yml` and then there would be no need to specify `--config` as the main entrypoint for the container, or for the readiness probe.

harshaisgud:
@joelddiaz Wow. That is an interesting find. But it still does not explain why the readiness probe is successful locally. The containers do go to a ready state locally without adding the path to the config file.

chris-rock:
Adding the path is correct. The question is more why it was even running without the provided path

.github/workflows/e2e.yaml
```diff
@@ -35,15 +38,29 @@ jobs:
...
+          echo "Wait for Operator"
+          kubectl wait --for=condition=Ready pod -l control-plane=controller-manager --namespace mondoo-operator
+          echo "display deployed pods"
+          kubectl describe pods -n mondoo-operator
+          echo "check mondoo status"
+          kubectl exec -q daemonset/mondoo-client --namespace mondoo-operator -- mondoo status
+          kubectl exec -q deployment/mondoo-client --namespace mondoo-operator -- mondoo status 
+          echo "wait for mondoo clients to be ready"
+          kubectl describe pods -n mondoo-operator 
+          kubectl wait --for=condition=Ready pod -l mondoo_cr=mondoo-client --namespace mondoo-operator --timeout=400s
```
chris-rock, line 53: It is important that timeout is longer than the configured PeriodSeconds since https://github.com/kubernetes/kubernetes/blob/047a6b9f861b2cc9dd2eea77da752ac398e7546f/pkg/kubelet/prober/worker.go#L134-L139 will set a runtime sleep time

joelddiaz:
In case someone else comes looking at the history here:
No matter the initialDelaySeconds value, if the kubelet has come up recently (and by "recently" we mean it has been running for less than the periodSeconds value on the probe), there will be a random sleep between 0 and periodSeconds before the probe starts running. With the original long period of 1800 seconds, and the fact that during an e2e run, the cluster has only been up for a moment as it is brought up for the express purpose of running e2e, we were seeing long delays before the Pods were considered Ready.

chris-rock (approved):
Thank you @harshaisgud @joelddiaz for solving that puzzle.

PR #236, opened 2022-03-30
increase olm timeout

joelddiaz:
Seeing timeout messages during e2e of the type:

time="2022-03-30T18:37:24Z" level=fatal msg="Failed to run bundle:
install plan is not available for the subscription
mondoo-operator-v0-2-1-sub: timed out waiting for the condition\n"

The default is 2m0s, so increase to 5m.


joelddiaz:
Tried running e2e steps locally and didn't have any issue. So try increasing timeout to address https://github.com/mondoohq/mondoo-operator/runs/5759513267?check_suite_focus=true#step:8:79

image

PR #241, opened 2022-03-31
WIP: add registry scanning fields for MondooAuditConfig

joelddiaz:
Extend MondooAuditConfig with a 'Registries' section.

It can hold a list of registry names/urls/endpoints that can be used to build a config to pass to the mondoo-client for scanning.
Rather than just a list of strings, each "registry" entry in the list is a distinct struct to allow more easily extending each entry with additional configuration fields in the future (things like scanning only specific repos, adding references to authentication for individual registries, etc).


api/v1alpha1/mondooauditconfig_types.go
```diff
@@ -52,6 +53,17 @@ type Image struct {
...
 
+// ContainerRegistries defines the configuration for container registry scanning.
+type ContainerRegistries struct {
+	Enable     bool       `json:"enable,omitempty"`
+	Registries []Registry `json:"registries,omitempty"`
+}
+
+// Registry holds all the data related to a single container image registry.
+type Registry struct {
+	Registry string `json:"registry"`
```
chris-rock, line 64: We also need to think about credentials. While public registries are great, we need to attach Kubernetes secrets in addition to the registries to be able to reach them.

We can do that in a second iteration.

chris-rock, line 64: We also need to think about credentials. While public registries are great, we need to attach Kubernetes secrets in addition to the registries to be able to reach them.

We can do that in a second iteration.

PR #247, opened 2022-04-05
🧹 Allowed mocking the function for retrieving the container image from a remote registry

### LINKED ISSUE #198: Allow stubbing/mocking out the image/tag fetching for test cases

imilchev:
Fixes #198

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -18,10 +18,15 @@ package controllers
 
 import (
 	"context"
+	"crypto/sha1"
+	"encoding/hex"
 	"path/filepath"
 	"testing"
 	"time"
 
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gexec"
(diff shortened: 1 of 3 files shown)


controllers/suite_test.go
```diff
@@ -55,6 +60,19 @@ func TestAPIs(t *testing.T) {
 var _ = BeforeSuite(func() {
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 
+	// Mock the retrieval of the actual image from the remote registry
+	getRemoteImage = func(ref name.Reference, options ...remote.Option) (*remote.Descriptor, error) {
```
joelddiaz, line 64: Seems so awkward to mock this out for these Gingko-style tests... But it does seem to work.

This kind of mock would also be needed for the non-Gingko-style tests in controlers/webhooks_test.go

imilchev: I will add the mock in the webhook tests as well. Just for traceability I will add what we discussed here.

We concluded that on the long term it will be best if we use the same testing style across all Mondoo projects. Therefore when we expanding the unit tests for the controller, we will gradually move away from Ginko-style tests.

joelddiaz (commented):
👍 LGTM

PR #248, opened 2022-04-05
 ⭐ Added test reporting for units tests Github workflow

imilchev:
The extra workflow is needed in order to be able to support PRs from forked repos. This is explained [here](https://github.com/marketplace/actions/test-reporter#recommended-setup-for-public-repositories).


joelddiaz:
I guess we just have to merge this before it starts working...

PR #251, opened 2022-04-05
⭐️ webhook mode of permissive/enforcing

### LINKED ISSUE #246: Allow defining enforcing/monitoring behavior for webhook
Add field under MondooAuditConfig.Spec.Webhook.Mode

Should be settable to "permissive" and "enforcing" with "permissive" as the default".

Pass this information to the Webhook (via env var?). To allow the webhook to either always allow items that go through the webhook through (and rely on the reporting into the Mondoo UI), or actually DENY the item creating/modification based on the results of the syncrounous scan.

joelddiaz:
- [x] Add new MondooAuditConfig.Spec.Webhooks.Mode that can be set to "";permissive;enforcing
- [x] Use the new Mode field to set MONDOO_WEBHOOK_MODE on the Deployment for the webhook
- [x] Test cases to cover a proper-looking webhook Deployment
- [x]  Random cleanups to make local testing easier

Not specific to #246, but still worth fixing:
- [x] Make sure we clean up Status when scanning is disabled for a specific component

closes #246

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -30,6 +30,7 @@ spec:
         args:
         - --leader-elect
         image: controller:latest
+        imagePullPolicy: IfNotPresent
         name: manager
         securityContext:
           allowPrivilegeEscalation: false
--- a/controllers/nodes.go
+++ b/controllers/nodes.go
@@ -155,7 +155,7 @@ func (n *Nodes) declareDaemonSet(ctx context.Context, clt client.Client, scheme 
 		}
 		return ctrl.Result{Requeue: true}, err
 	}
-	updateNodeConditions(n.Mondoo, found)
+	updateNodeConditions(n.Mondoo, found.Status.NumberReady != found.Status.DesiredNumberScheduled)
 	return ctrl.Result{}, nil
 }
 
@@ -315,6 +315,9 @@ func (n *Nodes) down(ctx context.Context, clt client.Client, req ctrl.Request) (
 		return ctrl.Result{}, err
 	}
 
+	// Update any remnant conditions
+	updateNodeConditions(n.Mondoo, false)
+
 	return ctrl.Result{Requeue: true}, err
 }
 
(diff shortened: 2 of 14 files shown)


pkg/webhooks/core/webhook.go
```diff
@@ -16,14 +19,31 @@ var corelog = logf.Log.WithName("core-validator")
 type CoreValidator struct {
 	Client  client.Client
 	decoder *admission.Decoder
+	Mode    string
```
imilchev, line 22: Isn't it better to hide this implementation behind an interface and have a function to initialize it. The function can also validate `Mode`, for example and return an error even before calling `Handle`. This would also allow for flexibility while testing. 

That is in case it is not an intentional choice that we only validate this every time `Handle` is called

config/manager/manager.yaml
```diff
@@ -30,6 +30,7 @@ spec:
         args:
         - --leader-elect
         image: controller:latest
+        imagePullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
joelddiaz: When set to Always, it means that even if I do a local build of the operator/container, k8s will try to pull down the image from a remote registry. And it was failing for me because I never actually pushed the locally built container out to docker hub.

controllers/nodes.go
```diff
@@ -315,6 +315,9 @@ func (n *Nodes) down(ctx context.Context, clt client.Client, req ctrl.Request) (
 		return ctrl.Result{}, err
 	}
 
+	// Update any remnant conditions
+	updateNodeConditions(n.Mondoo, false)
```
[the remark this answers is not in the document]
joelddiaz: Sounds good to me. Created https://github.com/mondoohq/mondoo-operator/issues/255 for this

pkg/webhooks/core/webhook.go
joelddiaz: What exactly do you have in mind here?

How would we enforce that someone didn't bypass the initialization and just built the CoreValidator directly?

imilchev: If we make the struct itself internal for this package, then externally you would be allowed to only instantiate it through the function that we create for it.

joelddiaz: I see what you mean. I changed it up to have a New() function and put some validation in that codepath so that the mode is checked/validated.

joelddiaz:
@imilchev this should be good for another review now

pkg/webhooks/core/webhook.go
```diff
@@ -16,14 +20,45 @@ var corelog = logf.Log.WithName("core-validator")
 type CoreValidator struct {
```
imilchev, line 20: if you want to make sure this cannot be created externally without using `NewCoreWebhook` you should make it internal `type coreValidator struct {` and also define a public interface for the validator itself. Then the `NewCoreWebhook` will return the interface instead of the `coreValidator` implementation.

We can discuss that verbally if that would make it easier.

pkg/webhooks/main.go
```diff
@@ -69,3 +84,12 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func validMode(mode string) bool {
```
imilchev, line 88: Is this actually used now?

joelddiaz: No. This is leftovers from when I was moving the mode checking logic around. I'll drop it in the next commit.

pkg/webhooks/core/webhook.go
joelddiaz: Most recent commit, should address your comments.

imilchev (approved):

imilchev:
LGTM

PR #254, opened 2022-04-06
🧹 cleanup and wait for the Pods to come up

joelddiaz:
Just like in the regular mondoo-namespace test, wait for the Pods. Otherwise we risk racing past to the next part of e2e and failing because we didn't wait.

More importantly, we've seen Pods fail to come up due to lack of resources in the minikube environment. So clean up the MondooAuditConfig from the mondoo-operator namespace test to free up some resource for the not-mondoo-operator tests.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -59,6 +59,8 @@ jobs:
           curl -sSL http://mondoo.io/download.sh | bash
           mv mondoo /usr/local/bin/ 
           make test/deployment
+          # Next test runs out of CPU resources if we leave these Pods running, so clean up after ourselves as we exit this test
+          kubectl delete -f config/samples/k8s_v1alpha1_mondooauditconfig.yaml --wait=true
       - name: Verify Mondoo Operator kustomize manifests in different namespace
         run: |
           kubectl create ns not-mondoo-operator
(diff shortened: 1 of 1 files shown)


joelddiaz:
Seeing the failures in the v0.2.2 branch, so we might as well merge this and rebase the 0.2.2 on top of this.

https://github.com/mondoohq/mondoo-operator/runs/5842250121?check_suite_focus=true#step:10:22

.github/workflows/e2e.yaml
```diff
@@ -69,13 +71,15 @@ jobs:
           yq ".metadata.namespace= \"not-mondoo-operator\" | .metadata.name = \"mondoo-operator-workload\"" config/rbac/workload_service_account.yaml  | kubectl apply -f -
           kubectl get clusterrolebinding mondoo-operator-workload -o yaml | yq '(.subjects) = [ .subjects[] + {"kind": "ServiceAccount", "name": "mondoo-operator-workload", "namespace": "not-mondoo-operator"}]' | kubectl apply -f -
           yq '.metadata.namespace= "not-mondoo-operator" | .spec.workloads.serviceAccount = "mondoo-operator-workload"' config/samples/k8s_v1alpha1_mondooauditconfig.yaml | kubectl apply -f -
+          sleep 20 # give minikube a moment to catch up with everything; else we get errors even when trying to wait for Pods if Pods simply don't yet exist
```
chris-rock, line 74: Can we avoid that by reducing the required resources? In general we have been pretty conservative.

chris-rock: I would like to avoid having a sleep 20 in because that will fall apart with slower machines and wait to long for faster machines. Can we run a wait kubectl command after the delete to make sure the deployment is gone?

chris-rock (approved):
Thank you for @joelddiaz I added some ideas we may want to do as followups

joelddiaz:
@chris-rock ok, I reworked it to avoid the explicit `sleep 20`. We now have our own homegrown `kubectl wait` using bash `until`. This gives us an approach like `kubectl wait` that doesn't error when the objects haven't yet appeared. This feels better than the random sleep 20.

I also created https://github.com/mondoohq/mondoo-operator/issues/256 to follow up with revisiting the resource requests/limits.

joelddiaz:
@imilchev PTAL and see if this approach makes sense

imilchev (approved):

imilchev:
LGTM! I just started working on migrating the e2e tests to Go, so this should be needed for just a while longer

PR #258, opened 2022-04-06
🧹 scale back the resource requests for Pods

### LINKED ISSUE #256: Restructure e2e test with resource contraints in mind
As discussed in https://github.com/mondoohq/mondoo-operator/pull/254#discussion_r843893158 , we should revisit the default resource requests/limits and see if they can be safely lowered.

joelddiaz:
As CPU is a very compressible resource, we can ask for far less than 375m (which is really basically saying we need 1/3 of a full CPU). Bring that way back to 50m (5% of a CPU). Still allowing the Pod access to a full 1/2 a CPU in the event that a particular scan was CPU intensive.

For Memory, scale it back a bit from requesting 750M (which was 75% of the Pod Memory limit), down to 500M (50% of the limit). This should guarantee a minimum of .5GB with a bit of "overscheduling" the underlying cluster memory-wise).

closes #256 

### DIFF
--- a/controllers/resources_requirements.go
+++ b/controllers/resources_requirements.go
@@ -11,10 +11,10 @@ var defaultMondooClientResources corev1.ResourceRequirements = corev1.ResourceRe
 		corev1.ResourceMemory: resource.MustParse("1G"),
 		corev1.ResourceCPU:    resource.MustParse("500m"),
 	},
-	// 75% of the limits
+
 	Requests: corev1.ResourceList{
-		corev1.ResourceMemory: resource.MustParse("750M"),
-		corev1.ResourceCPU:    resource.MustParse("375m"),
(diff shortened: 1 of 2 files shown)


joelddiaz:
Rebasing on top of #254 since without the waiting for the Pods, we still can just race by and fail later tests

Without the rebase:
```
namespace/not-mondoo-operator created
secret/mondoo-client created
serviceaccount/mondoo-operator-nodes created
serviceaccount/mondoo-operator-workload created
clusterrolebinding.rbac.authorization.k8s.io/mondoo-operator-workload configured
mondooauditconfig.k8s.mondoo.com/mondoo-client created
No resources found in not-mondoo-operator namespace.
```

imilchev (approved):

PR #259, opened 2022-04-06
⭐ Support operator-sdk 1.18

### LINKED ISSUE #245: Upgrade to operator-sdk 1.18.1
It appears that the first use of the operator-sdk started with 1.15 (based on the contents of files like https://github.com/mondoohq/mondoo-operator/blob/main/config/scorecard/patches/olm.config.yaml ) .

Walk through the published upgrade steps starting from 1.15 up through 1.18.1 ( https://sdk.operatorframework.io/docs/upgrading-sdk-version/ ) to ensure mondoo-operator is up-to-date with OperatorSDK 1.18.1

imilchev:
closes #245 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,39 +109,52 @@ docker-push: ## Push docker image with the manager.
 
 ##@ Deployment
 
+ifndef ignore-not-found
+  ignore-not-found = false
+endif
+
+.PHONY: install
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
 
+.PHONY: uninstall
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl delete -f -
+	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
+.PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -190,7 +203,7 @@ ifeq (,$(shell which opm 2>/dev/null))
 	set -e ;\
 	mkdir -p $(dir $(OPM)) ;\
 	OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
-	curl -sSLo $(OPM) https://github.com/operator-framework/operator-registry/releases/download/v1.15.1/$${OS}-$${ARCH}-opm ;\
+	curl -sSLo $(OPM) https://github.com/operator-framework/operator-registry/releases/download/v1.16.0/$${OS}-$${ARCH}-opm ;\
```
joelddiaz, line 206: Should this really be v1.16 even though we're migrating to OperatorSDK 1.18?

config/manager/manager.yaml
```diff
@@ -21,6 +21,8 @@ spec:
     metadata:
       labels:
         control-plane: controller-manager
+      annotations:
+        kubectl.kubernetes.io/default-container: manager
```
joelddiaz, line 25: Curious: Does this mean that for things like `kubectl logs` in a multi-container Pod, this will be auto-selected as the default one?

joelddiaz (commented):
Looks good. Just two questions.

Makefile
imilchev: It was not intended for review yet. I only implemented the changes for migrating from 1.15 to 1.16. I got stuck on how to manually verify that it still works. The good part is that the github workflow can validate it for me.

I think I will implement the changes for the migration between each version or in separate commits. Then we can see if somewhere it goes wrong

config/manager/manager.yaml
imilchev: Yes, I think that is the intention. However, I did not add these lines for that reason. This was just part of the instructions for the migration to the newer operator sdk version

joelddiaz (commented):
👍 LGTM

PR #260, opened 2022-04-06
WIP: just see what happens when a test fails

joelddiaz:
Just want to test out the new test reporting...


joelddiaz:
Seems to work just fine when the test cases fail!

Screen Shot 2022-04-06 at 2 45 09 PM

PR #267, opened 2022-04-11
🐛 Fixed RBAC definitions

### LINKED ISSUE #257: Don't create roles that grant permissions to "*" apiGroup
We have roles that need access to the core group (eg Pods, Secrets, etc), but the Role sets the APIGroup to "*". https://github.com/mondoohq/mondoo-operator/blob/main/config/rbac/role_nodes.yaml#L6

Should use the APIGroup of "" (the empty string) to indicate the core group.

imilchev:
closes #257 

### DIFF
--- a/controllers/mondooauditconfig_controller.go
+++ b/controllers/mondooauditconfig_controller.go
@@ -64,7 +64,6 @@ var deployInventoryyaml []byte
 //+kubebuilder:rbac:groups=cert-manager.io,resources=certificates;issuers,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=*,resources=*,verbs=get;list;watch
 //The last line is required as we cant assign higher permissions that exist for operator serviceaccount
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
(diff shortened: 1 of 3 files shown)


controllers/mondooauditconfig_controller.go
```diff
@@ -64,7 +64,6 @@ var deployInventoryyaml []byte
 //+kubebuilder:rbac:groups=cert-manager.io,resources=certificates;issuers,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=*,resources=*,verbs=get;list;watch
```
imilchev, line 67: I assumed that this is wrong as we already have specific permissions defined and then there is this global allow

joelddiaz, line 67: While we're here updating role permissions, I noticed that we are asking for create/delete/patch on MondooAuditConfig. the mondooauditconfig_controller responds to creation/deletion of MondooAuditConfig resources, but I'm not aware of any need that it has to create/delete/patch a MondooAuditConfig itself.

imilchev: I think the update is also irrelevant because we only update finalizers and status and those are separate rules

joelddiaz (commented):
👍 LGTM

joelddiaz:
For posterity: we presently need the mondooauditconfig_controller to have full Update permission on a MondooAuditConfig, because the RBAC rules that would normally allow just updating the finalizers list on the MondooAuditConfig are insufficient for how controller-runtime presently works. See https://github.com/kubernetes-sigs/kubebuilder/issues/2264

PR #268, opened 2022-04-11
🐛 make webhook names more unique

### LINKED ISSUE #252: Allow multiple MondooAuditConfigs with webhooks to be registred
Today, the name of the ValdiatingWebhookConfiguration is named in a way where there can be a name collision. 
MondooAuditConfig in namespace ABC named mondoo-client will have a ValidatingWebhookConfiguration named mondoo-client-mondoo-webhook

MondooAuditConfig in namespace XYZ named mondoo-client will have a ValidatingWebhookConfiguration named mondoo-client-mondoo-webhook

Add the namesapce into the generated name so that we avoid these collisions.

joelddiaz:
In the event that two MondooAuditConfigs with the same name in different Namespaces, the mondoo-operator would try to use the same name for both ValidatingWebhooks.

When generating the name for the webhook, use the Namespace in addition to the MondooAuditConfig name to avoid these kinds of naming colisions.

To avoid orphaning a previous Webhook using the old-style naming. Add some code to clean up the old-style-named webhook if it exists.

closes #252 

### DIFF
--- a/controllers/webhooks.go
+++ b/controllers/webhooks.go
@@ -82,9 +82,9 @@ func (n *Webhooks) syncValidatingWebhookConfiguration(ctx context.Context,
 	vwc *webhooksv1.ValidatingWebhookConfiguration,
 	annotationKey, annotationValue string) error {
 
-	// Override the default name to allow for multiple MondooAudicConfig resources
+	// Override the default/generic name to allow for multiple MondooAudicConfig resources
 	// to each have their own Webhook
-	vwc.Name = getValidatingWebhookName(n.Mondoo.Name)
+	vwc.Name = getValidatingWebhookName(n.TargetNamespace, n.Mondoo.Name)
 
 	// And update the Webhook entries to point to the right namespace/name for the Service
 	// receiving the webhook calls
@@ -131,7 +131,7 @@ func (n *Webhooks) syncValidatingWebhookConfiguration(ctx context.Context,
 		}
 	}
 
-	return nil
+	return n.cleanupOldWebhook(ctx)
 }
 
 func (n *Webhooks) syncWebhookService(ctx context.Context) error {
@@ -528,6 +528,10 @@ func (n *Webhooks) down(ctx context.Context) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
+	if err := n.cleanupOldWebhook(ctx); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	// Cleanup ValidatingWebhooks
 	r := bytes.NewReader(webhookManifestsyaml)
 	yamlDecoder := yamlutil.NewYAMLOrJSONDecoder(r, 4096)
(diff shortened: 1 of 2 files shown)


controllers/webhooks.go
```diff
@@ -611,8 +615,8 @@ func getWebhookDeploymentName(prefix string) string {
 	return prefix + "-webhook-manager"
 }
 
-func getValidatingWebhookName(prefix string) string {
-	return prefix + "-mondoo-webhook"
+func getValidatingWebhookName(mondooNamespace, mondooAuditConfigName string) string {
+	return fmt.Sprintf("%s-%s-mondoo", mondooNamespace, mondooAuditConfigName)
```
imilchev, line 619: shouldn't the last part be `-webhook` instead of `mondoo`? You could even use `-mondoo-webhook` but then it probably becomes kind of long.

The `mondooNamespace` should actually be the namespace of the audit config I guess? If that is the case, why not just have the audit config as a parameter and call the function something like `getWebhookNameForAuditConfig` (or make it shorter if you figure out a better name). We could even add the function to the audit config struct directly, so then you could do `auditConfig.ValidatingWebhookName()`.

controllers/webhooks.go
```diff
@@ -629,3 +633,21 @@ func updateWebhooksConditions(config *mondoov1alpha1.MondooAuditConfig, degraded
 	config.Status.Conditions = SetMondooAuditCondition(config.Status.Conditions, mondoov1alpha1.WebhookDegraded, status, reason, msg, updateCheck)
 
 }
+
+// TODO: This cleanup can be removed afer we are sure no more old-style
```
imilchev, line 637: on that note, do we have a plan on how to upgrade the operator. Let's say I have a version of the operator from 1 month ago and there are 2 newer versions. Do we aim to support upgrading directly to the latest or do we do it step-by-step (like Kubernetes, for example). This would determine for how long we have to keep temporary code such as this function.

controllers/webhooks.go
joelddiaz: I'm not a fan of adding the object type to the name of the object. It seems redundant to name a Secret `my-app-creds-secret` as the object is already of type `Secret`. You end up with command like `kubectl get secret my-app-creds-secret` which has the word `secret` in there too many times IMO.

I kept the `mondoo` part in the name so that someone listing all the webhooks in the cluster could quickly see which  webhooks were associated with Mondoo.

The name of the parameter `mondooNamespace` is misleading. The idea was that we would take the namespace of the MondooAuditConfig + the name of the MondooAuditConfig to build a unique Webhook name (and that is in fact what the code actually does). I'll fix the name, or just take in the MondooAuditConfig as a parameter instead.

Adding the function to the struct would make it more involved to reuse this function in the test cases.

controllers/webhooks.go
joelddiaz: Depending on how someone deployed mondoo-operator, we may have little control over how it was upgraded. For something like OLM, there is an idea of which release replaces the previous release, but for someone simply doing a `make deploy`, anything goes.

Until we have a more concrete understanding of how people will be deploying/upgrading mondoo-operator, we'll just have to be more cautious before removing these kinds of transition/temp functionality.

imilchev: well, we can define the manual here. For example, if you do `make deploy` you should do it incrementally. Maybe we should discuss this and see if we can already define a way of working

joelddiaz: Given our semantic versioning approach (so far), and that we are still in major version `0`, we are probably safe to remove this with version `0.3.x` (since we are in `0.2.x` today).

controllers/webhooks.go
```diff
@@ -615,8 +615,8 @@ func getWebhookDeploymentName(prefix string) string {
 	return prefix + "-webhook-manager"
 }
 
-func getValidatingWebhookName(mondooNamespace, mondooAuditConfigName string) string {
-	return fmt.Sprintf("%s-%s-mondoo", mondooNamespace, mondooAuditConfigName)
+func getValidatingWebhookName(mondooAuditConfigNamespace, mondooAuditConfigName string) string {
```
imilchev, line 618: I think it would be better if you just take the audit config as input, instead of using 2 parameters as you mentioned in your previous comment

imilchev (approved):

PR #269, opened 2022-04-11
🧹 descriptive names for DaemonSet and Deployment

### LINKED ISSUE #242: name DaemonSet and Deployment to reflect their purpose
Presently, the names of the DaemonSet and Deployment are just "mondoo-client". The Pods created for the DaemonSet/Deployment both have prefixes of "mondoo-client" which makes it unclear which Pod is doing what.

Name the DaemonSet/Deployments so that their purpose is clear and the Pods are more easily distinguishable. Additionally, include the name of the MondooAuditConfig into the names for these objects to allow multiple MondooAuditConfigs to exist in the same Namespace.
 
DaemonSet name should be: MONDOO_AUDIT_CONFIG_NAME-node
Deployment for workloads/api scanning should be: MONDOO_AUDIT_CONFIG_NAME-api

~~Lastly, update the name of the Deployment for the webhook to also include MONDOO_AUDIT_CONFIG_NAME as the prefix.~~ (Covered as part of #252 )

Be sure to clean up any previous Deployment/DaemonSet to cover the case of someone upgrading from a previous version of the mondoo-operator to this new naming scheme so that we don't leak resources.

joelddiaz:
Add the postfix '-node' to the DaemonSet so that the Pods can be more easily identified by name.

Add the postfix '-workload' to the Deployment (for workloads) so that the Pod can be more easily identified.

closes #242

### DIFF
--- a/controllers/nodes.go
+++ b/controllers/nodes.go
@@ -35,7 +35,7 @@ import (
 
 const (
 	daemonSetConfigMapNameTemplate = `%s-ds`
-	daemonSetNameTemplate          = `%s-node`
+	NodeDaemonSetNameTemplate      = `%s-node`
 )
 
 type Nodes struct {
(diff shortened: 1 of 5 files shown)


imilchev:
this needs a rebase as well

imilchev (approved):

PR #273, opened 2022-04-12
🧹 manually remove MondooAuditConfig for helm e2e

joelddiaz:
Just like we manually add the MondooAuditConfig, add a step to remove it before doing 'helm uninstall'.

Otherwise, the MondooAuditConfig resource doesn't get cleaned up because the finalizer sticks around, and because it doesn't get removed, then the node/workload scanning resources also stick around.


chris-rock (approved):
Thank you @joelddiaz

PR #274, opened 2022-04-12
🧹 make resource comments reflect reality

joelddiaz:
Recently updated the resource requests/limits (in #258 ). Test case had old comment about what the percentages were. Make comments in test reflect reality.

Hat tip @tas50 


chris-rock (approved):
Thank you @joelddiaz

PR #275, opened 2022-04-13
🧹 Added comment to explain RBAC configuration

imilchev:
I forgot to add this comment for the RBAC. It's good to have it so we don't change it again later 


joelddiaz (commented):
👍 LGTM

PR #276, opened 2022-04-13
🧹 Re-use finalizer functions from controller-runtime

### LINKED ISSUE #271: Re-use finalizer functions from controller-runtime
We should re-work the finalizer logic to actually use the functions from [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/b8db76e66383819c94f530bd4b8dea60025b716f/pkg/controller/controllerutil/controllerutil.go) instead of re-implementing it ourselves.

imilchev:
closes #271 


joelddiaz (commented):
👍 LGTM

PR #279, opened 2022-04-13
:star: Add linting step

imilchev:
I noticed there is no linting step for the code base (only a spell checker). This PR adds [golangci-lint](https://golangci-lint.run/) action to the repo and extends the `Makefile` with a lint target

### DIFF
--- a/tests/integration/mondoo_installation_test.go
+++ b/tests/integration/mondoo_installation_test.go
@@ -23,18 +23,12 @@ func (s *MondooInstallationSuite) SetupSuite() {
 	s.ctx = context.Background()
 }
 
-func (s *MondooInstallationSuite) TearDownSuite() {
-	if s.testCluster != nil {
-		s.testCluster.UninstallOperator()
-	}
-}
-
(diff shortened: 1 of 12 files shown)


tests/integration/mondoo_installation_test.go
```diff
@@ -111,7 +105,11 @@ func (s *MondooInstallationSuite) testMondooInstallation() {
 func TestMondooInstallationSuite(t *testing.T) {
 	s := new(MondooInstallationSuite)
 	defer func(s *MondooInstallationSuite) {
-		HandlePanics(recover(), s.TearDownSuite, s.T)
+		HandlePanics(recover(), func() {
+			if err := s.testCluster.UninstallOperator(); err != nil {
+				zap.S().Errorf("Failed to uninstall Mondoo operator. %v", err)
+			}
+		}, s.T)
```
joelddiaz, line 112: The prevous TearDownSuite() function had some checks around s.testCluster not being nil. Have we lost that, or is that check no longer necessary due to some other changes I have overlooked?

joelddiaz:
I think this just needs a rebase.

PR #283, opened 2022-04-14
🐛 Integration tests for audit config in non-default namespace

### LINKED ISSUE #280: restore e2e test for non-default namespace
The original e2e test for non-default namespaces was set up to run the mondoo-operator in the `mondoo-operator` namespace, and create a MondooAuditConfig in a Namespace that wasn't `mondoo-operator`.

With the implementation of the new go-based e2e tests, the non-default test case actually runs the mondoo-operator and the MondooAuditConfig in the `some-namespace` Namespace.

While running the mondoo-operator itself in a different Namespace is also a useful thing to be able to validate, we should restore the test where mondoo-operator runs in one Namespace and the MondooAuditConfig is created in a different namespace.

imilchev:
Closes #280 

Also removed service account, roles and role bindings that were related to node scanning since they are not actually needed.


tests/framework/installer/installer.go
```diff
@@ -91,6 +82,25 @@ func (i *MondooInstaller) UninstallOperator() error {
...
+	secret := &corev1.Secret{}
+	secret.Name = utils.MondooClientSecret
+	secret.Namespace = i.Settings.Namespace
+	if err := i.K8sHelper.DeleteResourceIfExists(secret); err != nil {
+		return err
+	}
+
+	_, err := i.K8sHelper.KubectlWithStdin(i.readManifestWithNamespace(OperatorManifest), utils.DeleteIngoreNotFoundFromStdinArgs...)
+	if err != nil {
+		return fmt.Errorf("Failed to delete mondoo-operator pod: %v ", err)
```
joelddiaz, line 98:
```suggestion
		return fmt.Errorf("Failed to delete mondoo-operator manifest(s): %v ", err)
```

tests/integration/mondoo_installation_test.go
```diff
@@ -7,9 +7,12 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/suite"
+	mondoov1 "go.mondoo.com/mondoo-operator/api/v1alpha1"
```
joelddiaz, line 10: I would group this import down with the other mondoo-operator packages below

tests/integration/mondoo_installation_test.go
```diff
@@ -19,40 +22,76 @@ import (
 
 type MondooInstallationSuite struct {
 	suite.Suite
-	ctx         context.Context
-	testCluster *TestCluster
+	ctx          context.Context
+	testCluster  *TestCluster
+	objToCleanup []client.Object
```
joelddiaz, line 27:
```suggestion
	objsToCleanup []client.Object
```

controllers/nodes.go
```diff
@@ -197,6 +201,7 @@ func (n *Nodes) daemonsetForMondoo() *appsv1.DaemonSet {
 						Key:    "node-role.kubernetes.io/master",
 						Effect: corev1.TaintEffect("NoSchedule"),
 					}},
+					AutomountServiceAccountToken: &automountServiceAccountToken,
```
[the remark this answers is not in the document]
imilchev: TIL

imilchev:
should be good to go now!

joelddiaz (approved):
👍

PR #284, opened 2022-04-14
:star: extend webhook watching to Deployments

### LINKED ISSUE #249: Extend webhooks to watch for Deployments
Add a new webhook to watch for Deployments (since Deployments are not part of the core API and we cannot simply extend the existing Webhook entrypoint to accept Deployments).

Copy the existing pkg/webhooks/core/* into pkg/webhooks/apps and ensure the comment to generate the Webhook manifests tells controller-gen to watch for Deployments (from the appsv1 group).

Then register the new "endpoint" in pkg/webhooks/main.go

joelddiaz:
- [x] use existing webhook to also watch for appsv1/Deployments
- [x] rename files to reflect more generic handling of resources
- [x] register new endpoint for receiving appsv1 resources
- [x] generate updated manifests

Additionally, implement a custom deepEquals() for knowing when an existing webhook needs updating.

On update of an existing webhook, be sure to clear out any previous annotations to cover the case where a runtime change was made to handing TLS certs for the webhook.

closes #249

### DIFF
--- a/controllers/webhook-manifests.yaml
+++ b/controllers/webhook-manifests.yaml
@@ -16,25 +16,6 @@ webhooks:
   rules:
   - apiGroups:
     - ""
-    apiVersions:
-    - v1
-    operations:
-    - CREATE
-    - UPDATE
-    resources:
-    - pods
-  sideEffects: None
-- admissionReviewVersions:
-  - v1
-  clientConfig:
-    service:
-      name: webhook-service
-      namespace: system
-      path: /validate-k8s-mondoo-com-apps
-  failurePolicy: Ignore
-  name: apps-policy.k8s.mondoo.com
-  rules:
-  - apiGroups:
     - apps
     apiVersions:
     - v1
@@ -42,5 +23,6 @@ webhooks:
     - CREATE
     - UPDATE
     resources:
+    - pods
     - deployments
   sideEffects: None
--- a/controllers/webhook-manifests.yaml
+++ b/controllers/webhook-manifests.yaml
@@ -10,9 +10,9 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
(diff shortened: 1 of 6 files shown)


controllers/webhook-manifests.yaml
```diff
@@ -24,3 +24,23 @@ webhooks:
...
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validate-k8s-mondoo-com-apps
+  failurePolicy: Ignore
+  name: apps-policy.k8s.mondoo.com
+  rules:
+  - apiGroups:
```
imilchev, line 37: can't these be combined into 1 webhook or is there some limitation that won't let us do it?

joelddiaz: It turns out that you can do it. I'll push a commit with exactly that change in a moment.

pkg/webhooks/main.go
```diff
@@ -68,7 +68,6 @@ func main() {
 		os.Exit(1)
 	}
 	hookServer.Register("/validate-k8s-mondoo-com-core", &webhook.Admission{Handler: webhookValidator})
```
imilchev, line 70: do we still want to call this `/validate-k8s-mondoo-com-core`? Maybe now it can be just `/validate-k8s-mondoo-com`?

joelddiaz: I looked into changing it. It can certainly be done as a new version of mondoo-operator will come with a new version of the webhook. We just don't presently have a way to know that a re-deployment of the Webhook Deployment is necessary. So I built and launched the new mondoo-operator, but the webhook deployment didn't change.
Let me go ahead an change it since it does bother me a little to have the old name in there...

imilchev (approved):
lgtm

PR #285, opened 2022-04-14
⭐ Deploy scan API

### LINKED ISSUE #277: Deploy Mondoo scan API when webhook is enabled
Create a Kubernetes deployment that runs the Mondoo scan API and a Kubernetes service in front of it. This is the prerequisite for performing scans in the admission controller

imilchev:
- [x] Create a deployment for the scan API
- [x] Always check if the deployment needs to be updated
- [x] Create a service for the scan API
- [x] Always check if the service needs to be updated
- [x] Split off `Version` variable from webhook implementation
- [x] Print version in console when starting the operator and the webhook
- [ ] ~~Unit tests~~ <- will follow up in a separate PR so we do not block each other
- [x] Integration tests

I also want to see if I can re-structure a little the implementation, to make it easier to read. Let's see how that will go.

Closes #277


controllers/scan_api_utils.go
```diff
@@ -0,0 +1,126 @@
...
+	"go.mondoo.com/mondoo-operator/api/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	scanApiDeploymentName = "mondoo-scan-api"
+	scanApiServiceName    = "mondoo-scan-api"
```
joelddiaz, line 15: These names should be created based on the name of the MondooAuditConfig to allow multiple MondooAuditConfigs in the same Namespace to each have their own ScanAPI Deployment.

controllers/webhooks.go
```diff
@@ -504,6 +502,38 @@ func (n *Webhooks) Reconcile(ctx context.Context) (ctrl.Result, error) {
 	return ctrl.Result{}, nil
 }
 
+func (n *Webhooks) deployScanApi(ctx context.Context) error {
```
joelddiaz, line 505: Organizationally, this function should probably exist in the scan_api_utils.go file as one day the node/workload scanning codepaths will want to make use of this.

controllers/scan_api_utils.go
```diff
@@ -0,0 +1,126 @@
...
+	scanApiPort           = 8080
+)
+
+var scanApiLabels = map[string]string{"app": "mondoo-scan-api"}
+
+func ScanApiDeployment(ns string, m *v1alpha1.MondooAuditConfig) appsv1.Deployment {
+	return appsv1.Deployment{
```
joelddiaz, line 22:
```suggestion
func ScanApiDeployment(ns string, m *v1alpha1.MondooAuditConfig) *appsv1.Deployment {
	return &appsv1.Deployment{
```

controllers/scan_api_utils.go
```diff
@@ -0,0 +1,126 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      scanApiDeploymentName,
```
joelddiaz, line 24: Some kind of templating here, but at the very least:
```suggestion
			Name:      fmt.Sprintf("%s-scan-api", m.Name),
```

controllers/scan_api_utils.go
```diff
@@ -0,0 +1,126 @@
...
+			Selector: &metav1.LabelSelector{
+				MatchLabels: scanApiLabels,
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: scanApiLabels,
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{
+						Image:     mondooImage,
```
joelddiaz, line 38: Following the pattern of how the node/workload (and webhook) config allow you to override the image used. We probably want a way to override the image/tag for the (future) shared ScanAPI container. We can probably delay on deciding how we want to expose the ability to override this image to a later PR.

But we do want to resolve this image to the container SHA (which will allow "seamless" upgrades as new versions of the mondoo-client are published and the mondooauditconfig_controller.Reconcile() re-fetches/updates the SHA over time). All this image resolution should be done while respecting the value of MondooOperatorConfig.Spec.SkipContainerResolution.

controllers/scan_api_utils.go
```diff
@@ -0,0 +1,126 @@
...
+						Name:      "mondoo-client",
+						Command:   []string{"mondoo", "serve", "--api", "--config", "/etc/opt/mondoo/mondoo.yml"},
+						Resources: getResourcesRequirements(m.Spec.Workloads.Resources),
```
joelddiaz, line 41: Since we don't prsently expose controlling the values for the ScanAPI Deployment, perhaps just taking the default is appropriate here.

controllers/scan_api_utils.go
```diff
@@ -0,0 +1,126 @@
...
+						Ports: []corev1.ContainerPort{
+							{ContainerPort: scanApiPort, Protocol: corev1.ProtocolTCP},
+						},
+						Env: []corev1.EnvVar{
+							{Name: "DEBUG", Value: "false"},
+							{Name: "MONDOO_PROCFS", Value: "on"},
+							{Name: "PORT", Value: fmt.Sprintf("%d", scanApiPort)},
+						},
+					}},
+					ServiceAccountName: m.Spec.Workloads.ServiceAccount,
```
joelddiaz, line 80: For anything but the `mondoo-operator` Namespace, the ServiceAccount would need to be defined. It is odd to have to specify fields in .spec.workload.* to affect the Webhook / ScanAPI. So perhaps we will need to figure out how to expose controls over the ScanAPI Deployment in this PR afterall...

controllers/scan_api_utils.go
```diff
@@ -0,0 +1,126 @@
...
+		},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{
+				{
+					Port:       int32(scanApiPort),
+					Protocol:   corev1.ProtocolTCP,
+					TargetPort: intstr.FromInt(scanApiPort),
+				},
+			},
+			Selector: scanApiLabels,
```
joelddiaz, line 123: The labels for the Deployment would also need to be unique enough to allow more than one ScanAPI Deployment to exist in the same Namespace.

controllers/utils.go
```diff
@@ -111,6 +112,18 @@ func parseReference(log logr.Logger, container string) (string, error) {
 	return imageUrl, nil
 }
 
+// AreDeploymentsDifferent returns a value indicating whether 2 deployments are different. Note that it does not perform a full
+// comparison but checks just some of the properties of a deployment (only the ones we are currently interested at).
+func AreDeploymentsDifferent(a, b appsv1.Deployment) bool {
```
joelddiaz, line 117:
```suggestion
func AreDeploymentsDifferent(a, b *appsv1.Deployment) bool {
```

joelddiaz (commented):
Pretty nice how easy the mondoo-client is to stand up in "serve" mode 🥇 

Depending on how the functionality progresses, we may need to deal with defining a place in MondooAuditConfig.Spec to control the details about the ScanAPI configuration.

controllers/scan_api_utils.go
imilchev: I thought that we need just 1 scan API for the entire cluster regardless of the config

controllers/scan_api_utils.go
imilchev: why return the pointer in this case? Now that I think about it, the `MondooAuditConfig` should probably be a reference instead of a pointer as well. This is because we are not modifying it in that function.

controllers/scan_api_utils.go
joelddiaz: Because the scanAPI runs with mondoo credentials, there is the possibility that you set up a MondooAuditConfig using your creds, and I set up a MondooAuditConfig with my creds. So in this case, they would each run with their own creds and fetch policies/upload results into their own Mondoo space.

controllers/scan_api_utils.go
joelddiaz: Not sure I understand the question. My suggestion was to return a pointer to a Deployment because what I really care about is the underlying client.Object that lets you call things like .DeepCopy()/.GetOwnerReference()/etc on the object.

controllers/scan_api_utils.go
imilchev: I would, indeed, prefer to split that in a separate PR. This one is already becoming quite big. No need to make it even bigger

controllers/utils.go
imilchev: I am intentionally avoiding using pointers in functions that do not and are not supposed to modify the structs. I think in this way it's easy to see if a function makes changes or it's just reading the struct. By having the interface like this it also makes sure that if someone decides to start modifying your parameters inside that function, the changes will not result in weird behaviour up in the call stack

controllers/scan_api_utils.go
imilchev: I will add those. I guess that also means that the webhook deployment itself has the same issue? The labels there are just static as well

controllers/scan_api_utils.go
imilchev: do we actually need the service account for doing scans? I thought that the idea is that the webhook will be sending the YAML over to the scan API. If that is the case, then the service account is not needed at all?

imilchev:
This should be good for review now

controllers/scan_api_utils.go
joelddiaz: If I understand the question correctly: For the ScanAPI to handle the payloads from the webhook, the ServiceAccount wouldn't' be needed as the mondoo client would have the full k8s resource already. For future capabilities where we route workload scanning through the ScanAPI, then the ServiceAccount would be needed for the ScanAPI to talk to k8s directly.
My expectation is that in the future the MondooAuditConfig will expose a new (for example) .Spec.ScanAPI.* set of fields to affect things related to the ScanAPI Deployment resource.

tests/integration/audit_config_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (s *AuditConfigSuite) TestReconcile_Nodes() {
+	auditConfig := utils.DefaultAuditConfig(s.testCluster.Settings.Namespace, false, true, false)
+	s.testMondooAuditConfigNodes(auditConfig)
+}
+
+func (s *AuditConfigSuite) TestReconcile_Webhooks() {
+	auditConfig := utils.DefaultAuditConfig(s.testCluster.Settings.Namespace, false, false, true)
+	s.testMondooAuditConfigWebhooks(auditConfig)
+}
```
joelddiaz, line 29: 🥇

controllers/webhooks/webhooks.go
```diff
@@ -102,6 +103,10 @@ func (n *Webhooks) syncValidatingWebhookConfiguration(ctx context.Context,
 	for i := range vwc.Webhooks {
 		vwc.Webhooks[i].ClientConfig.Service.Name = getWebhookServiceName(n.Mondoo.Name)
 		vwc.Webhooks[i].ClientConfig.Service.Namespace = n.Mondoo.Namespace
+
+		if vwc.Webhooks[i].ClientConfig.Service.Port == nil {
+			vwc.Webhooks[i].ClientConfig.Service.Port = pointer.Int32(443)
+		}
```
joelddiaz, line 83: We can probably be safe just making this port 443 setting a static value/codepath. Either we control that field or we don't. And since the ValidatingWebhookConfiguration resource is created by us, we get to decide what that field should contain at creation/update time.

joelddiaz (commented):
Looks great. Some questions/comments

controllers/scanapi/resources.go
```diff
@@ -0,0 +1,160 @@
...
+			Replicas: pointer.Int32(1),
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{
+						Image:     image,
+						Name:      "mondoo-client",
+						Command:   []string{"mondoo", "serve", "--api", "--config", "/etc/opt/mondoo/mondoo.yml"},
```
joelddiaz, line 58: One thing I ran into when testing the scan api locally (not through the Deployment) was that I had to set the `--token` parameter for my scans to work. And whatever I set as the `--token` would be required to match from my local scan URL calls. So we probably will need some kind of system to generate the token and sync the token between the Deployment and the Webhook. Maybe not strictly necessary for this PR, but I figured I'd write down what I ran into for when it comes time to do the final linking of the webhook to the ScanAPI.

controllers/webhooks.go
```diff
@@ -597,20 +535,13 @@ func (n *Webhooks) down(ctx context.Context) (ctrl.Result, error) {
...
-	scanApiDeployment := ScanApiDeployment(n.TargetNamespace, n.Mondoo)
-	if err := genericDelete(ctx, n.KubeClient, &scanApiDeployment); err != nil {
-		webhookLog.Error(err, "failed to clean up scan API Deployment resource")
-		return ctrl.Result{}, err
-	}
-
-	scanApiService := ScanApiService(n.TargetNamespace)
-	if err := genericDelete(ctx, n.KubeClient, &scanApiService); err != nil {
-		webhookLog.Error(err, "failed to clean up scan API Service resource")
+	if err := scanapi.Cleanup(ctx, n.KubeClient, n.TargetNamespace, *n.Mondoo); err != nil {
```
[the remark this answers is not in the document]
imilchev: Indeed, we should extract it at a later point. At least the actual code that performs the actions is already isolated

controllers/scanapi/resources.go
imilchev: I guess we just try it out. I used the command that was mentioned by Chris in his PR of the main repo.

controllers/scanapi/deploy.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		logger.Error(err, "Failed to create Deployment for scan API")
+		return err
+	}
+
+	if created {
+		logger.Info("Created Deployment for scan API")
+	} else if !k8s.AreDeploymentsEqual(deployment, existingDeployment) {
+		// If the deployment exists but it is different from what we actually want it to be, then update.
+		if err := kubeClient.Update(ctx, &deployment); err != nil {
```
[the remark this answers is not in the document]
imilchev: good question, I haven't tried that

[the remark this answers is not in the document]
imilchev: I made sure we override only specific fields

controllers/webhooks/webhooks.go
imilchev: yeah, right now it's empty but Kubernetes puts a default of 443. I added this so at least we do not override the service constantly because of the comparison failing due to default values

imilchev: This approach at least makes sure that if we manually change the validation webhook manifest, the code will not override the port.

joelddiaz (approved):
👍 LGTM

PR #289, opened 2022-04-15
⭐ e2e for webhook

### LINKED ISSUE #287: Create an e2e test for deploying admission with a custom certificate

joelddiaz:
Add some steps to test out the webhook functionality.

Manually create the certificates/Secret.
Enable the webhook.
Convert the webhook to a FailurePolicy of Fail (so that we can show the webhook is active and failing because we haven't completed the certificate configuration).
Attempt to modify a Deployment and expect it to fail (b/c Cert config not complete).
Add the CA data to the webhook.
Attempt to modify a Deployment and expect it to succeed.

closes #287 

### DIFF
--- a/tests/framework/utils/certificate_utils.go
+++ b/tests/framework/utils/certificate_utils.go
@@ -1,0 +1,78 @@
+package utils
+
+import (
+	"bytes"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"time"
+
+	"github.com/stretchr/testify/suite"
+)
+
+// Webhook generation approach from https://gist.github.com/velotiotech/2e0cfd15043513d253cad7c9126d2026#file-initcontainer_main-go
+
+// GenerateWebhookCerts will return the CA certificate, the Server certificate, and the Server private key
+func GenerateWebhookCerts(s suite.Suite, dnsNames []string) (*bytes.Buffer, *bytes.Buffer, *bytes.Buffer) {
+	ca := &x509.Certificate{
+		SerialNumber: big.NewInt(2222),
+		Subject: pkix.Name{
+			Organization: []string{"Mondoo Operator E2E CA"},
+		},
+		NotBefore: time.Now(),
+		NotAfter:  time.Now().AddDate(0, 0, 5),
+		IsCA:      true,
+		//ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		BasicConstraintsValid: true,
+	}
+
+	caPrivKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	s.NoError(err, "Failed to generate private key for CA")
+
+	caCert, err := x509.CreateCertificate(rand.Reader, ca, ca, &caPrivKey.PublicKey, caPrivKey)
+	s.NoError(err, "Failed to create self-sign certificate for CA")
+
+	pemEncodedCA := new(bytes.Buffer)
+	s.NoError(pem.Encode(pemEncodedCA, &pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: caCert,
+	}))
+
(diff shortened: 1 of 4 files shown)


tests/framework/utils/certificate_utils.go
```diff
@@ -0,0 +1,78 @@
...
+	"time"
+
+	"github.com/stretchr/testify/suite"
+)
+
+// Webhook generation approach from https://gist.github.com/velotiotech/2e0cfd15043513d253cad7c9126d2026#file-initcontainer_main-go
+
+// GenerateWebhookCerts will return the CA certificate, the Server certificate, and the Server private key
+func GenerateWebhookCerts(s suite.Suite, dnsNames []string) (*bytes.Buffer, *bytes.Buffer, *bytes.Buffer) {
+	ca := &x509.Certificate{
```
imilchev, line 20: wow this is nice! My original plan was to just generate the files once manually and just re-use them. This is even better 🤩

tests/framework/utils/certificate_utils.go
```diff
@@ -0,0 +1,78 @@
...
+		SerialNumber: big.NewInt(2222),
+		Subject: pkix.Name{
+			Organization: []string{"Mondoo Operator E2E CA"},
+		},
+		NotBefore: time.Now(),
+		NotAfter:  time.Now().AddDate(0, 0, 5),
+		IsCA:      true,
+		//ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
```
imilchev, line 28: if this is not needed, then perhaps remove it?

tests/framework/utils/certificate_utils.go
```diff
@@ -0,0 +1,78 @@
...
+		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		BasicConstraintsValid: true,
+	}
+
+	caPrivKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	s.NoError(err, "Failed to generate private key for CA")
```
imilchev, line 34:
```suggestion
	s.NoErrorf(err, "Failed to generate private key for CA")
```

You should use `NoErrorf` everywhere you want to provide a specific error message. I saw that you used `NoError` in multiple places. I won't add a separate comment for each. Perhaps just go through the assertions once more and change where needed

tests/integration/mondoo_installation_test.go
```diff
@@ -147,6 +151,123 @@ func (s *MondooInstallationSuite) testMondooAuditConfig(auditConfig mondoov1.Mon
 	s.Equalf(1, len(daemonSets.Items), "DaemonSets count in Mondoo namespace is incorrect.")
 	expectedDaemonSetName := fmt.Sprintf(mondoocontrollers.NodeDaemonSetNameTemplate, auditConfig.Name)
 	s.Equalf(expectedDaemonSetName, daemonSets.Items[0].Name, "DaemonSet name does not match expected name based from audit config name.")
+
+	zap.S().Info("Enable webhook.")
+
+	// Generate certificates manually
+	serviceDNSNames := []string{
```
imilchev, line 158: I would extract the creation of the secret to a separate function in some helper file. Maybe in `installer.go`? We might need to re-use this. It will also make this test specification easier to read if we just call one function `CreateWebhookCertSecret` (or something similar).

imilchev (changes requested):

tests/integration/mondoo_installation_test.go
```diff
@@ -160,114 +160,102 @@ func (s *MondooInstallationSuite) testMondooAuditConfig(auditConfig mondoov1.Mon
...
-		},
-	}
-	s.Require().NoError(
-		s.testCluster.K8sHelper.Clientset.Get(s.ctx, client.ObjectKeyFromObject(operatorConfig), operatorConfig),
-		"Failed to get existing MondooOperatorConfig")
+	secretName := "webhook-server-cert"
+	caCert, err := s.testCluster.MondooInstaller.GenerateServiceCerts(&auditConfig, secretName, serviceDNSNames)
+
+	// Don't bother with further webhook tests if we couldnt' save the certificates
+	if s.NoErrorf(err, "Error while generating/saving certificates for webhook service") {
```
imilchev, line 167: isn't it sufficient to do `s.Require().NoErrorf(...)` in this case to stop the test in case this block fails?

PR #291, opened 2022-04-18
🐛 generate Webhook Secret names from MondooAuditConfig

### LINKED ISSUE #290: generate unique Secret names for Webhook Deployment
Two MondooAuditConfigs in the same Namespace that enable Webhooks will try to use the same Secret name for their certificate data.

While this may be useful for an environment where sharing the Secret data is doable, this approach precludes someone from having an environment where distinct certificates are desired.

Generate a Secret name based on the name of the MondooAuditConfig resource so that each Webhook Deployment can mount its own unique Secret.

joelddiaz:
Previously, the Secret holding the TLS certificates for the webhook pod was statically named. This meant that two MondooAuditConfigs in the same Namespace that both wanted to enable webhooks would have a colision on the name of the Secret holding the certificates.

Use the MondooAuditConfig name to built up the name of the Secret to allow multiple MondooAuditConfigs with webhooks enabled in the same Namespace.

closes #290


joelddiaz:
This PR builds on top of the changes in #289 , so hold off on merging until that one merges first.

joelddiaz:
> This PR builds on top of the changes in #289 , so hold off on merging until that one merges first.

#289 merged, so rebased this on main, and this PR should be ready for review/merging.

imilchev (approved):

PR #300, opened 2022-04-21
🐛 Allow node scanning on rancher-provisioned controlplane and etcd nodes

### LINKED ISSUE #288: Node scanning for Rancher-provisioned clusters
I tried deploying the operator to a Rancher-provisioned cluster on a few virtual machines. The node scanning did not get scheduled on the master nodes because at the moment the DaemonSet tolerates only the `node-role.kubernetes.io/master` taint. However, in Rancher there are 2 different taints that are applied on the nodes depending on which components are running on them - namely:
- [node-role.kubernetes.io/etcd](https://rancher.com/docs/rke/latest/en/config-options/nodes/#etcd)
- [node-role.kubernetes.io/controlplane](https://rancher.com/docs/rke/latest/en/config-options/nodes/#controlplane)

If we would like the operator to support Rancher-provisioned clusters, we should make sure these labels are added to the list of tolerations.

imilchev:
Closes #288 

I tested it manually to verify these changes are sufficient to fix the issue. It seems to be working!

### DIFF
--- a/controllers/nodes.go
+++ b/controllers/nodes.go
@@ -210,7 +210,7 @@ func (n *Nodes) daemonsetForMondoo() *appsv1.DaemonSet {
 							Value:  "true",
 						},
 						{
-							// Rancher controlplaen node
+							// Rancher controlplane node
 							// https://rancher.com/docs/rke/latest/en/config-options/nodes/#controlplane
 							Key:    "node-role.kubernetes.io/controlplane",
 							Effect: corev1.TaintEffectNoSchedule,
--- a/controllers/nodes.go
+++ b/controllers/nodes.go
@@ -197,10 +197,22 @@ func (n *Nodes) daemonsetForMondoo() *appsv1.DaemonSet {
 					Labels: ls,
 				},
 				Spec: corev1.PodSpec{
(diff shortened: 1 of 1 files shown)


joelddiaz (commented):
This looks perfectly fine to me, but can we put comments near each of the tolerations that explains what environment each entry specifically enables?

imilchev:
I added comments explaining the tolerations

controllers/nodes.go
```diff
@@ -197,10 +197,26 @@ func (n *Nodes) daemonsetForMondoo() *appsv1.DaemonSet {
...
+						},
+						{
+							// Rancher etcd node
+							// https://rancher.com/docs/rke/latest/en/config-options/nodes/#etcd
+							Key:    "node-role.kubernetes.io/etcd",
+							Effect: corev1.TaintEffectNoExecute,
+							Value:  "true",
+						},
+						{
+							// Rancher controlplaen node
```
joelddiaz, line 213:
```suggestion
							// Rancher controlplane node
```

joelddiaz (approved):
👍

PR #301, opened 2022-04-22
⭐ Cache resolved container images per reconcile cycle

### LINKED ISSUE #296: Refactor container image resolving logic
Currently, the container image resolution is performed multiple times for a single `Reconcile` cycle. When the controller reconciles the `MondooAuditConfig` for each separate section (workloads, nodes, webhooks) the container images are being resolved.

It might be best if we create a struct that when initialized will resolve all images (operator and client). We can then pass that struct to each of the separate section. To make it even better for testing, we can hide that entire logic behind an interface such as:
```go
type ContainerImageResolver interface {
  MondooOperatorImage() string
  MondooClientImage() string
}
```

The actual implementation of this interface can even perform lazy loading to make sure we do not resolve images that are not actually needed.

imilchev:
- [x] Implement a `ContainerImageResolver` to cache resolved container images per reconcile cycle.
- [x] Implement a "fake" `ContainerImageResolver` to be used when testing such that no remote calls are performed during unit tests
- [x] Unit test the `ContainerImageResolver` implementation

Closes #296


imilchev:
This should be good for a review

joelddiaz (approved):
👍 This is a huge improvement.

Maybe in the future we can have longer-lived caching (maybe attached to the MondooAuditConfig.Status??) to be resilient against short-lived container registry outages.

PR #303, opened 2022-04-22
🌟 Scan on webhook handling

### LINKED ISSUE #278: Use scan API to scan resources in admission controller
Use the scan API as deployed in #277 to perform scans in the admission controller

joelddiaz:
Building on top of the work from #285 :

- [x] Create token Secret for Scan API to use as auth (add Role permissions for Secret Create/Delete)
- [x] Update Scan API Deployment to use Secret/token
- [x] Pass webhook token info (via Secret) and URL to call via new webhook CLI parameters
- [x] Stop using env var for setting webhook mode, and just use another CLI parameter
- [x] Start calling Scan API when receiving a webhook
- [x] Send labels with the scan call 
- [x] Update scanner tests to be usable in webhook tests

Unrelated changes to how the namespace detection is implemented (thought it would be used in the webhook, but ended up not needing it).

closes #278


controllers/scanapi/deploy.go
```diff
@@ -57,6 +66,25 @@ func Cleanup(ctx context.Context, kubeClient client.Client, ns string, mondoo mo
 	return nil
 }
 
+func createSecret(ctx context.Context, kubeClient client.Client, mondoo mondoov1alpha1.MondooAuditConfig) error {
+	scanApiTokenSecret := ScanApiSecret(mondoo)
+	if err := ctrl.SetControllerReference(&mondoo, scanApiTokenSecret, kubeClient.Scheme()); err != nil {
+		return err
+	}
+
+	err := kubeClient.Create(ctx, scanApiTokenSecret)
```
imilchev, line 75: I think it would be better if you use `k8s.CreateIfNotExist` unless there is some super specific reason not to use it.

controllers/scanapi/resources.go
```diff
@@ -19,21 +19,41 @@ package scanapi
...
 	DeploymentSuffix = "-scan-api"
 	ServiceSuffix    = "-scan-api"
+	SecretSuffix     = "-scan-api-token"
 	Port             = 8080
 )
 
+func ScanApiSecret(mondoo v1alpha1.MondooAuditConfig) *corev1.Secret {
+
+	// Generate a token. It will only be saved on initial Secret creation.
+	token := uuid.New()
```
imilchev, line 44: not related to the code but what is the purpose of this `token` exactly?

controllers/scanapi/deploy.go
```diff
@@ -34,6 +35,9 @@ var logger = ctrl.Log.WithName("scan-api-deploy")
 // Deploy deploys the scan API for a given MondooAuditConfig. The function checks if the scan API is already deployed.
 // If that is the case, the existing resources are compared with the ones that are desired and the necessary updates are applied.
 func Deploy(ctx context.Context, kubeClient client.Client, ns, image string, mondoo mondoov1alpha1.MondooAuditConfig) error {
+	if err := createSecret(ctx, kubeClient, mondoo); err != nil {
```
imilchev, line 38: make sure you extend the tests to verify the secret as well

pkg/utils/k8s/equality.go
```diff
@@ -15,8 +15,10 @@ func AreDeploymentsEqual(a, b appsv1.Deployment) bool {
 		reflect.DeepEqual(a.Spec.Selector, b.Spec.Selector) &&
 		reflect.DeepEqual(a.Spec.Template.Spec.Containers[0].Image, b.Spec.Template.Spec.Containers[0].Image) &&
 		reflect.DeepEqual(a.Spec.Template.Spec.Containers[0].Command, b.Spec.Template.Spec.Containers[0].Command) &&
+		reflect.DeepEqual(a.Spec.Template.Spec.Containers[0].Args, b.Spec.Template.Spec.Containers[0].Args) &&
```
imilchev, line 18: extend the tests for these changes

pkg/webhooks/handler/webhook.go
```diff
@@ -34,31 +40,73 @@ func NewWebhookValidator(client client.Client, mode string) (admission.Handler,
...
+	}
+
+	handlerlog.Info("Scan result", "shouldAdmit", passed, "kind", req.Kind.Kind, "resource", resource)
 
 	// Depending on the mode, we either just allow the resource through no matter the scan result
 	// or allow/deny based on the scan result
 	switch a.mode {
 	case mondoov1alpha1.Permissive:
-		return admission.Allowed("PASSED")
+		response = admission.Allowed("PASSED MONDOO SCAN")
```
imilchev, line 98: Not sure where exactly this info is visible but don't we want to mention that the scan failed but the pod is still allowed because we are in permissive mode?

pkg/webhooks/handler/webhook.go
```diff
@@ -67,3 +115,37 @@ func (a *webhookValidator) InjectDecoder(d *admission.Decoder) error {
 	a.decoder = d
 	return nil
 }
+
+type customObjectMeta struct {
```
imilchev, line 119: perhaps add a comment why this is needed and for what it is used

pkg/webhooks/handler/webhook.go
```diff
@@ -67,3 +115,37 @@ func (a *webhookValidator) InjectDecoder(d *admission.Decoder) error {
...
+	objMeta := &customObjectMeta{}
+
+	yamlDecoder := yamlutil.NewYAMLOrJSONDecoder(r, 4096)
+
+	if err := yamlDecoder.Decode(objMeta); err != nil {
+		return nil, err
+	}
+
+	labels := map[string]string{
+		"k8s.mondoo.com/namespace": objMeta.Namespace,
```
imilchev, line 136: maybe define these as consts so it's easy to find them and re-use them in tests?

pkg/webhooks/handler/webhook.go
```diff
@@ -67,3 +115,37 @@ func (a *webhookValidator) InjectDecoder(d *admission.Decoder) error {
...
+		"k8s.mondoo.com/uid":       string(objMeta.UID),
+		"k8s.mondoo.com/name":      objMeta.Name,
+		"k8s.mondoo.com/kind":      objMeta.Kind,
+	}
+
+	for _, or := range objMeta.GetOwnerReferences() {
```
imilchev, line 142: You can use [this](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/controller_ref.go#L33) 😛

pkg/webhooks/handler/webhook_test.go
```diff
@@ -14,36 +14,50 @@ import (
...
+			}
+
+			testserver := fakescanapi.FakeServer()
 			validator := &webhookValidator{
 				decoder: decoder,
+				mode:    test.mode,
+				scanner: &scanner.Scanner{
+					Endpoint: testserver.URL,
+				},
 			}
```
imilchev, line 61: I think it's really worth testing the `generateLabels` function because it is kind of critical to what we are trying to achieve

pkg/webhooks/main.go
```diff
@@ -39,6 +43,31 @@ func init() {
 func main() {
 	webhookLog := log.Log.WithName("webhook")
 
+	var scanURL string
+	var webhookMode string
+	var tokenFilePath string
+	flag.StringVar(&scanURL, "scan-url", "", "The URL of the Service to send scan requests to.")
```
imilchev, line 49: maybe call it `scan-api-url` and then also adjust the description to make it clear that this is the URL for the scan API

pkg/webhooks/main.go
```diff
@@ -39,6 +43,31 @@ func init() {
...
+	flag.StringVar(&tokenFilePath, "token-file-path", "", "Path to file containing token to use when making scan requests.")
+	flag.StringVar(&webhookMode, "enforcement-mode", string(mondoov1alpha1.Permissive), "What action to take on a failed scan.")
```
imilchev, line 51: I would make the description of this a little more elaborate on what the 2 modes do and how they differ

tests/integration/audit_config_base_suite.go
```diff
@@ -104,7 +104,7 @@ func (s *AuditConfigBaseSuite) testMondooAuditConfigNodes(auditConfig mondoov1.M
 func (s *AuditConfigBaseSuite) testMondooAuditConfigWebhooks(auditConfig mondoov1.MondooAuditConfig) {
```
imilchev, line 104: perhaps it's good to add a new test suite to validate the webhook itself. For example, try to create a few resources in the different modes and assert the outcome. Could be done in a separate PR as well but I think it's important that we address this right after finishing the scan API integration

controllers/scanapi/resources.go
joelddiaz: The client when in scan mode will take a --token flag, and actual ScanAPI() calls to the Service require the client making the scan request to have a matching token.
So we just need to generate something that the ScanAPI Deployment and the Webhook Deployment can both synchronize on.

pkg/webhooks/handler/webhook.go
joelddiaz: We can change the text for Permissive + failed scan to something like `PERMITTING FAILED SCAN`. WDYT?

imilchev: Yes, that's what I meant. In this way you can also see (wherever this is visible) that you scan failed but we are still allowing because of your admission settings

joelddiaz: I will constify these strings, and put some descriptive comments around when each is applicable.

controllers/scanapi/deploy.go
imilchev: it's a legit reason, maybe write a comment about it so we don't forget

controllers/scanapi/deploy_test.go
```diff
@@ -37,12 +39,23 @@ func (s *DeploySuite) TestDeploy_Create() {
 	ns := "test-ns"
 	auditConfig := utils.DefaultAuditConfig(ns, false, false, true)
 
-	client := fake.NewClientBuilder().WithScheme(s.scheme).Build()
+	kubeClient := fake.NewClientBuilder().WithScheme(s.scheme).Build()
 
-	s.NoError(Deploy(s.ctx, client, ns, s.image, auditConfig))
+	s.NoError(Deploy(s.ctx, kubeClient, ns, s.image, auditConfig))
```
imilchev, line 44: I guess the cleanup should be extended as well to verify that the secret is deleted

imilchev (approved):
LGTM

PR #305, opened 2022-04-25
⭐ Defined v1alpha2

### LINKED ISSUE #286: Rename webhook in MondooAuditConfig to admission
The `webhook` section of the MondooAuditConfig should be renamed to `admission` for clarity

imilchev:
Closes #286 


api/v1alpha2/mondooauditconfig_types.go
```diff
@@ -0,0 +1,161 @@
...
+
+package v1alpha2
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// MondooAuditConfigSpec defines the desired state of MondooAuditConfig
+type MondooAuditConfigData struct {
```
joelddiaz, line 25: Calling it `Data` has always struck me as a weird thing. I'm used to seeing `.*Spec` here in CRD definitions.
```suggestion
type MondooAuditConfigSpec struct {
```

api/v1alpha2/mondooauditconfig_types.go
```diff
@@ -0,0 +1,161 @@
...
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Config is an example field of MondooAuditConfig. Edit mondooauditconfig_types.go to remove/update
+	CertificateProvisioning CertificateProvisioning `json:"certificateProvisioning,omitempty"`
```
joelddiaz, line 29: Does this make more sense to exist under the `Admission` section, since these values only affect how the Webhook is deployed?

api/v1alpha2/mondooauditconfig_types.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+type Admission struct {
+	Enable bool  `json:"enable,omitempty"`
+	Image  Image `json:"image,omitempty"`
+	// Mode represents whether the webhook will behave in a "permissive" mode (the default) which
+	// will only scan and report on k8s resources or "enforcing" mode where depending
+	// on the scan results may reject the k8s resource creation/modification.
+	// +kubebuilder:validation:Enum="";permissive;enforcing
+	Mode AdmissionMode `json:"mode,omitempty"`
```
joelddiaz, line 67: Let's do the same defaulting as is done for the CertificateProvisioning.Mode field.

joelddiaz (commented):
As I looked through this, I couldn't help but wonder whether we should introduce v1alpha2 alongside moving node/workload scanning to run under the new ScanAPI. Moving those things under ScanAPI may also justify some additional changes to the MondooAuditConfig API.

imilchev:
> As I looked through this, I couldn't help but wonder whether we should introduce v1alpha2 alongside moving node/workload scanning to run under the new ScanAPI. Moving those things under ScanAPI may also justify some additional changes to the MondooAuditConfig API.

This is what we discussed with Chris. Let's assume that this is the CRD we will use for the scan API scanning. So let's consider this as step 1 in that process.

I appreciate you looking at this PR at this early stage but it is far from finished and might change a lot still. I didn't want to waste your time with checking unfinished code so that's why a created it as a draft. Don't feel obliged to look at the drafts though. Check them only if you have spare time. Otherwise I will ping you when it's ready for review 🙏

api/v1alpha1/mondooauditconfig_types.go
```diff
@@ -147,6 +149,84 @@ type MondooAuditConfig struct {
...
+	dst.Spec.Admission.Image.Tag = src.Spec.Webhooks.Image.Tag
+
+	// TODO: add status
+
+	return nil
+}
+
+// ConvertFrom converts from the Hub version (v1alpha2) to this version.
+func (dst *MondooAuditConfig) ConvertFrom(srcRaw conversion.Hub) error {
+	src := srcRaw.(*v1alpha2.MondooAuditConfig)
```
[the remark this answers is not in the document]
imilchev: I took that from the example of kubebuilder. They basically guarantee that this line should always succeed.

api/v1alpha2/mondooauditconfig_types.go
imilchev: yup

docs/user-manual.md
```diff
@@ -0,0 +1,327 @@
...
+We recently upgraded our CRDs version to `v1alpha2` and currently manual migration steps need to be executed. You can list the CRDs with the old version by running:
+```bash
+kubectl get mondooauditconfigs.v1alpha1.k8s.mondoo.com -A
+```
+
+Each of the CRDs in the list needs to be manually edited and mapped to the new version. As an example, to edit a CRD called `mondoo-client` in namespace `mondoo-operator` the following command should be executed:
+```bash
+kubectl edit mondooauditconfigs.v1alpha1.k8s.mondoo.com mondoo-client -n mondoo-operator
+```
+
```
imilchev, line 326: I intentionally did not add the specific steps to convert the version in text. Not sure what is the best way to describe it and also not even sure if it is needed at all. I guess this should suffice for now

config/certmanager/certificate.yaml
```diff
@@ -0,0 +1,25 @@
+# The following manifests contain a self-signed issuer CR and a certificate CR.
```
joelddiaz, line 1: What is the purpose of these config/certmanager/* files?

config/crd/kustomization.yaml
```diff
@@ -9,7 +9,7 @@ resources:
 patchesStrategicMerge:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
 # patches here are for enabling the conversion webhook for each CRD
-#- patches/webhook_in_mondooauditconfigs.yaml
+- patches/webhook_in_mondooauditconfigs.yaml
```
joelddiaz, line 12: Since we don't actually deploy the webhook, is this necessary?

controllers/nodes.go
```diff
@@ -148,7 +145,7 @@ func (n *Nodes) declareDaemonSet(ctx context.Context, clt client.Client, scheme
 	}
 
 	// check that the resource limites are identical
-	expectedResourceRequirements := k8s.ResourcesRequirementsWithDefaults(n.Mondoo.Spec.Nodes.Resources)
+	expectedResourceRequirements := k8s.ResourcesRequirementsWithDefaults(n.Mondoo.Spec.Scanner.Resources)
```
joelddiaz, line 148: This is cool. It gives us a nice off-ramp to a day where we can just use the ScanAPI completely.

docs/deployment-configurations.md
```diff
@@ -6,35 +6,8 @@ Mondoo Client can be configured to run security policies against Kubernetes obje
...
-          - id: api
-            connections:
-              - backend: k8s
-                options:
-                  namespace: "example"
-  nodes:
-    enable: true
-  mondooSecretRef: mondoo-client
-```
+~~To run Mondoo policies against Kubernetes objects in the example namespace the following configuration can be used.~~
```
joelddiaz, line 9: We can also just show an example MondooAuditConfig showing the capabilities that presently *are* available

docs/user-manual.md
```diff
@@ -0,0 +1,327 @@
...
+2. Apply the configuration via:
+    ```bash
+    kubectl apply -f mondoo-config.yaml
+    ```
+
+## Deploying the Admission controller
+K8s webhooks require TLS certs to establish the trust between the Certificate Authority listed in the `ValidatingWebhookConfiguration.Webhooks[].ClientConfig.CABundle` and the TLS certificates presented when connecting to the HTTPS endpoint specified in the webhook.
+
+You can choose to install and use cert-manager to automate the creation and updating of the TLS certs, or you can create (and rotate) your own TLS certificates manually.
```
joelddiaz, line 116:
```suggestion
You can choose to install and use cert-manager to automate the creation and updating of the TLS certs, or you can use the OpenShift certificate creation/rotation features, or you can create (and rotate) your own TLS certificates manually.
```

docs/user-manual.md
```diff
@@ -0,0 +1,327 @@
...
+            name: mondoo-operator-webhook-service
+            namespace: mondoo-operator
+            path: /validate-k8s-mondoo-com-core
+            port: 443
+    ```
+
+## Uninstalling operator
+Before uninstalling the operator make sure all `MondooAuditConfig` and `MondooOperatorConfig` objects are deleted. You can verify if there are any of them in your cluster by running:
+```bash
+kubectl get mondooauditconfigs.k8s.mondoo.com,mondoooperatorconfigs.k8s.mondoo.com -A
```
joelddiaz, line 244: There's probably no need to delete the MondooOperatorConfig beforehand as it doesn't have a finalizer to block things from getting automatically cleaned up.

api/v1alpha1/mondooauditconfig_types.go
```diff
@@ -137,6 +139,8 @@ const (
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:deprecatedversion
+//+kubebuilder:deprecatedversion:warning="k8s.mondoo.com/v1alpha1 is deprecated. The CRD will be automatically converted to v1alpha2"
```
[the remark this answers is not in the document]
imilchev: that's correct. It was a copy-pasta thing. The `MondooOperatorConfig` will be automatically converted but the `MondooAuditConfig` won't.

api/v1alpha1/mondooauditconfig_types.go
```diff
@@ -147,6 +151,87 @@ type MondooAuditConfig struct {
...
+func (src *MondooAuditConfig) ConvertTo(dstRaw conversion.Hub) error {
+	dst := dstRaw.(*v1alpha2.MondooAuditConfig)
+
+	dst.ObjectMeta = *src.ObjectMeta.DeepCopy()
+
+	dst.Spec.Admission.CertificateProvisioning.Mode =
+		v1alpha2.CertificateProvisioningMode(src.Spec.Webhooks.CertificateConfig.InjectionStyle)
+	if dst.Spec.Admission.CertificateProvisioning.Mode == "" {
+		dst.Spec.Admission.CertificateProvisioning.Mode = v1alpha2.ManualProvisioning
+	}
```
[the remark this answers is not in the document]
imilchev: Not sure and since I cannot test it now and we will be referring to this piece of code for upgrade steps, I made it explicit

config/certmanager/certificate.yaml
imilchev: auto-generated by kubebuilder. If you enable them in `kustomization.yaml` they will make sure your webhook is getting a certificate via cert-manager automatically. I have all of them disabled at the moment. The individual files are still being generated though and that's why they are in the repo

config/crd/kustomization.yaml
imilchev: yes, it is. This particular patch patches the CRD such that it specifies conversion via a webhook. If the patch is disabled then the CRD will have conversion `None` and will do the "dumb" conversion by just dropping fields.

docs/deployment-configurations.md
imilchev: I intentionally left it there as strike-through text. The `inventory` is currently disabled as we discussed with Chris that currently we are unsure how we want to work with it. I think it's nice to have that visible in the docs that it is something we do not support now.

Also we agreed last week to spend some time on sorting out the documentation for the operator. I guess it's nice to have the info here and let Letha and Tim decide what to do with it 😝

docs/user-manual.md
imilchev: it doesn't have a finalizer for now that's true. But I left it for consistency and to be sure that even if we decide to have a finalized on the operator config, it will still work. I mean, it doesn't hurt if the operator config is deleted as well I guess

imilchev:
I renamed `webhooks` to `admission` to make the changes consistent and also renamed the integration tests.

I noticed that our integration tests do not have a validation for the `MondooAuditConfig` status updated so I added an issue for that #312 .

api/v1alpha2/mondooauditconfig_types.go
```diff
@@ -0,0 +1,163 @@
...
+type CertificateProvisioning struct {
+	// +kubebuilder:validation:Enum=cert-manager;openshift;manual
+	// +kubebuilder:default=manual
+	Mode CertificateProvisioningMode `json:"mode,omitempty"`
+}
+
+// Scanner defines the settings for the Mondoo scanner that will be running in the cluster. The same scanner
+// is used for scanning the Kubernetes API, the nodes and for serving the admission controller.
+type Scanner struct {
+	MondooCredsSecretRef string                      `json:"mondooCredsSecretRef,omitempty"`
```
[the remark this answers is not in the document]
imilchev: I moved it to the root

imilchev:
I added a default value for `ServiceAccountName` this means that also the check whether it is specified or not in combination with the non-default namespace is now gone. There will be no explicit message for this but it will fail to create the workload. If we want to add a specific message for this case again, I suggest we add a separate issue for that

controllers/admission/admission_handler.go
```diff
@@ -48,7 +48,7 @@ var (
 	webhookManifestsyaml []byte
 )
 
-type Webhooks struct {
+type AdmissionHandler struct {
```
joelddiaz, line 51: AdmissionHandler seems like an odd name since this isn't actually the code that handles the webhook(s).

api/v1alpha2/mondooauditconfig_types.go
```diff
@@ -43,6 +43,7 @@ type CertificateProvisioning struct {
 // Scanner defines the settings for the Mondoo scanner that will be running in the cluster. The same scanner
 // is used for scanning the Kubernetes API, the nodes and for serving the admission controller.
 type Scanner struct {
+	// +kubebuilder:default=mondoo-operator-workload
```
joelddiaz, line 48: As we're in the middle of retiring the use of "workload" to describe the "k8s resource" scanning, we probably should have this name reflect its purpose. Perhaps `mondoo-k8s-api`?

Makefile
```diff
@@ -104,10 +104,10 @@ test/ci: manifests generate fmt vet envtest gotestsum
 
 # Integration tests are run synchronously to avoid race conditions
 test/integration: manifests generate generate-manifests load-minikube
-	go test -v -p 1 ./tests/integration/...
+	go test -v -timeout 975s -p 1 ./tests/integration/...
```
joelddiaz, line 107: 975 seems like such an oddly specific number. Why 975?

api/v1alpha2/mondooauditconfig_types.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+// MondooAuditConfigSpec defines the desired state of MondooAuditConfig
+type MondooAuditConfigSpec struct {
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Config is an example field of MondooAuditConfig. Edit mondooauditconfig_types.go to remove/update
+	// +kubebuilder:validation:Required
+	// +required
+	MondooCredsSecretRef string              `json:"mondooCredsSecretRef"`
```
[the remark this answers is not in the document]
imilchev: I would avoid linking it to an object from the core API if we can make it self-contained

api/v1alpha2/mondooauditconfig_types.go
imilchev: I didn't change that because that would mean there will be a dangling service account left behind. I guess that's fine. Better change the name now than later. I will adjust the code

Makefile
imilchev: brain freeze. Should be 900

controllers/admission/admission_handler.go
imilchev: `AdmissionDeploymentHandler` should fit better

joelddiaz (approved):

PR #309, opened 2022-04-28
🧹 Speed up container image build process

### LINKED ISSUE #219: Cache go modules and images
To speed up the CI we need some caching of the build dependencies. The lowest hanging fruit seem to be the go modules.

Cached dependencies on the github runner

imilchev:
- Added caching for Go modules and Go build cache.
  - Speeds up building, unit and integration testing. 
  - Saves ~2mins for each of these steps
- Build release binaries outside of the container and only copy them in a container for the specific platform. 
  - Makes sure we do not rely on emulation for building containers for different architectures
  - Saves >1h for release build

Closes #219


imilchev:
I verified that the resulting containers are correctly signed and do work on all the architectures we actually push. One thing I could not verify is whether this setup works with multiple tags. I built it having this in mind but it is not tested.

For example, when building for a tag on `main` 2 tags are being pushed to the container registry - the actual tag from the git tag and `latest`. Both need to be pushed and signed. We should keep an eye on that during the next release to verify that it works.

Makefile
```diff
@@ -108,15 +112,16 @@ test/integration/ci: manifests generate generate-manifests load-minikube gotests
...
-	go build -o bin/webhook pkg/webhooks/main.go
+	CGO_ENABLED=0 GOOS=$(TARGET_OS) GOARCH=$(TARGET_ARCH) go build -o bin/manager -ldflags "-s -w -X go.mondoo.com/mondoo-operator/pkg/version.Version=${VERSION}" main.go
+	CGO_ENABLED=0 GOOS=$(TARGET_OS) GOARCH=$(TARGET_ARCH) go build -o bin/webhook pkg/webhooks/main.go
 
 run: manifests generate fmt vet ## Run a controller from your host.
 	MONDOO_OPERATOR_NAMESPACE=mondoo-operator go run ./main.go
 
-docker-build: ## Build docker image with the manager.
-	docker build --build-arg VERSION=${VERSION} -t ${IMG} .
+docker-build: build ## Build docker image with the manager.
```
joelddiaz, line 121: Does it ever make sense to do a docker-build with a TARGET_OS != linux? Basically, should this make target also set TARGET_OS=linux statically just like load-minikube?

Dockerfile
```diff
@@ -1,31 +1,9 @@
...
-COPY main.go main.go
-COPY api/ api/
-COPY controllers/ controllers/
-COPY pkg/ pkg/
-
-# Build
-ARG VERSION
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH go build -a -o manager -ldflags "-X go.mondoo.com/mondoo-operator/pkg/version.Version=$VERSION" main.go
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH go build -a -o webhook pkg/webhooks/main.go
-
```
joelddiaz, line 22: IIUC, we're basically giving up on doing the go binary builds in a container environment, and we'll use the ubuntu-based github runner image as our "builder image"?

scripts/push-virtual-tag.sh
```diff
@@ -0,0 +1,37 @@
+# exit immediately when a command fails
+set -e
+# only exit with zero if all commands of the pipeline exit successfully
+set -o pipefail
+
+export COSIGN_EXPERIMENTAL=true
+export DOCKER_CLI_EXPERIMENTAL=enabled
```
joelddiaz, line 7: I guess worst case here is that the API of the 'docker manifest' CLI changes, and we adjust. I can't imagine that they will abandon 'docker manifest' completely in favor of 'buildx' 🤞...

Dockerfile
imilchev: yes, this is the thing that was actually making our build incredibly slow. We can build Go perfectly fine for (almost) any target architecture natively. Instead we were building it in a container. For the different architectures that container had to be emulated because the entire container is running under a different arch and that's why it took 1h to get all the images

scripts/push-virtual-tag.sh
imilchev: maybe that is the case. I did take this approach from `prometheus-operator` so I guess it should be stable-enough for now 😝

joelddiaz (approved):
👍 I'm looking forward to this faster build process on Tuesday :)

PR #320, opened 2022-05-03
e2e-olm: gather logs from all pods generically

joelddiaz:
Since the mondoo-operator Deployment pod isn't coming up, we aren't getting useful pod logs.

Try gathering logs from all Pods in the mondoo-operator namespace if something failed.


imilchev:
You might be better off adding that thing that allows SSH access to the runner to debug this.

imilchev (approved):

joelddiaz:
> You might be better off adding that thing that allows SSH access to the runner to debug this.

Yes, that is on my todo list as well :) I guess now that the builds happen so much faster it isn't unreasonable to just watch the build and then SSH in after failure within an hour.

PR #323, opened 2022-05-05
⭐ token to service account

### LINKED ISSUE #310: have mondoo-operator take time-limited tokens and generate a mondoo Service Account
Allow for the possibility to provide a JWT token from the Mondoo console as part of the MondooAuditConfig creation.

Extend MondooAuditConfig to include .spec.mondooTokenRef to hold a reference to a Secret containing the token data.

Secret.data['token'] will hold the token data.

Choice 1:
Early in the mondooauditconfig_controller, when .spec.mondooTokenRef is non-empty and .spec.mondooSecretRef points to a non-existent Secret, have the mondoo-operator generate the Mondoo service account using the ExchangeRegistrationToken API and save it into the SecretRef specified in .spec.mondooSecretRef.

This requires a combination of creating the Secret with the token data, and populating the name of the Secret where mondoo-operator should save the service account data.

Choice 2:
Allow .spec.mondooTokenRef to be non-empty and .spec.mondooSecretRef to be empty.

mondoo-operator updates the MondooAuditConfig.spec.mondooSecretRef to point to the just-created Secret. This is a bit unusual (but not unheard of) to have the controller update the .spec fields for an object that the controller is watching/processing.

This can complicate the process of applying "updates" to the MondooAuditConfig as the MondooAuditConfig data retrived from install.mondoo.com may not know the name of the Secret the mondoo-operator created (unless we agree to a known pattern to use for the Secret name).

Choice 3:
If .spec.mondooTokenRef is non-empty, save the Secret using a predictable naming pattern.

Then when the Mondoo creds are needed, the code can use the .spec.mondooSecretRef (for the case where the user brought long-lived service account Secret with them), or fall back to checking for the existence of the predictably named Secret used to hold the generated service account data.

This avoid having mondoo-operator be responsible for writing to MondooAuditConfig.spec. And we can just say that .spec.mondooSecretRef takes precedence over any generated Secret.

More choices???

For any of these choices, when things aren't configured properly (ie. .spec.mondooTokenRef is empty and .spec.mondooSecretRef is empty for 'Choice 1' or .spec.mondooTokenRef is non-empty but .spec.mondooSecretRef is populated for 'Choice 2'), we should attach a status.condition to the MondooAuditConfig to communicate what is wrong.

joelddiaz:
Allow specifying a MondooAuditConfig.Spec.MondooTokenSecretRef. If the Secret referenced in .Spec.MondooCredsSecretRef doesn't exists, and MondooTokenSecretRef does exist, use the token data to exchange for a Mondoo service account and save it into the MondooCredsSecretRef Secret.

Create a new mondooclient package (modeled after the scanner package) to wrap the interactions with the Mondoo API. Generate a mock client to allow testing of the various token/creds scenarios.

Add RBAC to the mondoo-operator ClusterRole to allow reading the Secrets named 'mondoo-client' and 'mondoo-token'. A user wishing to specify different Secret names will need to update the RBAC rules to allow mondoo-operator to read those alternative Secret names.

Note: Using the kubernetes client-go client to do the Secret checking because controller-runtime will try to List() all Secrets.

closes #310


controllers/mondooauditconfig_controller.go
```diff
@@ -56,6 +64,11 @@ var (
 
 	// Defined as a var so we can easily mock this in tests.
 	createContainerImageResolver = mondoo.NewContainerImageResolver
+
+	// so we can mock out the mondoo client for testing
+	MondooClientBuilder = func(opts mondooclient.ClientOptions) mondooclient.Client {
```
imilchev, line 69:
```suggestion
	MondooClientBuilder = mondooclient.NewClient
```

I think this is equivalent but then shorter

controllers/mondooauditconfig_controller.go
```diff
@@ -251,6 +275,107 @@ func getPodNames(pods []corev1.Pod) sets.String {
...
+	}
+
+	// mondoCredsExists is already false from here down
+	if !mondooTokenExists {
+		log.Info("neither .spec.MondooCredsSecretRef nor .spec.MondooTokenSecretRef exist")
+		return nil
+	}
+
+	log.Info("Creating Mondoo service account from token")
+	token := strings.TrimSuffix(string(mondooTokenSecret.Data["token"]), "\n")
```
imilchev, line 320: you can use `strings.TrimSpaces(string(mondooTokenSecret.Data["token"]))`. That covers spaces, tabs and new lines

imilchev (approved):
time to ship it :shipit:

PR #334, opened 2022-05-11
Mondoo console integration

### LINKED ISSUE #326: Add k8s integration support
Extend MondooAuditConfig to allow specifying that a specific MondooAuditConfig is to be handled as a full "k8s integration" with the Mondoo API.

When enabled, an alternative token-to-service-account path should be taken to ensure the "integration" is properly registered.

Add a periodic controller (running every 10 minutes) to wake up and walk through all MondooAuditConfigs with the integration enabled, and "phone home" to the Mondoo API (using the previously generated Mondoo service account in .spec.MondooCredsSecretRef) to report on the health of the integration.

Read through the MondooAuditConfig.Status to determine the health state of the integration, and pass along the full MondooAuditConfig as part of the status report.

joelddiaz:
- [x] Introduce MondooAuditConfig.Spec.ConsoleIntegration.Enable. This will allow indicating that the mondoo-operator should periodically CheckIn() to the Mondoo API on behalf of the k8s integration that was created via the Mondoo console.
- [x] When consoleIntegration is enabled, and exchanging a token for a service account, use the Register() API call instead of ExchangeRegistrationToken() (this means we expect the 'owner' to be set on the token claims so we know the MRN of the integration). When saving the Mondoo service account in the standard 'config' Secret key, also save the integration MRN in the same Secret under the 'integrationmrn' key.
- [x] New 'integration' controller which runs every 10 minutes. It lists all MondooAuditConfigs, and for the ones with consoleIntegration.Enabled perform a CheckIn().
- [x] Test cases to cover the various posibilities when doing a CheckIn()

closes #326 


controllers/integration/integration_controller.go
```diff
@@ -0,0 +1,181 @@
...
+
+	zapLog, err := cfg.Build()
+	if err != nil {
+		return fmt.Errorf("failed to set up logging for integration controller: %s", err)
+	}
+	log = zapr.NewLogger(zapLog)
+
+	mc := &IntegrationReconciler{
+		Client:              mgr.GetClient(),
+		Interval:            10 * time.Minute,
```
imilchev, line 47: can we extract that as a `const` so we can easily find it in case we need to change it in the future

controllers/integration/integration_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	Client client.Client
+
+	// Interval is the length of time we sleep between runs
+	Interval            time.Duration
+	log                 logr.Logger
+	mondooClientBuilder func(mondooclient.ClientOptions) mondooclient.Client
+}
+
+// Start begins the integration status loop.
+func (r *IntegrationReconciler) Start(ctx context.Context) error {
```
imilchev, line 68: I think it will be best if this context is stored in the struct and is then used for all calls that are made in `integrationLoop` below. This will make sure that in case there are pending calls and we need to exit, those calls will be canceled.

controllers/integration/integration_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	switch pk := key.(type) {
+	case *ecdsa.PrivateKey:
+		return r.buildTokenFromPrivateKey(pk, serviceAccount)
+	default:
+		return "", fmt.Errorf("AuthKey must be of type ecdsa.PrivateKey")
+	}
+}
+
+func (r *IntegrationReconciler) buildTokenFromPrivateKey(pk *ecdsa.PrivateKey, sa *mondooclient.ServiceAccountCredentials) (string, error) {
```
imilchev, line 161: the name of this function is a bit strange. It is creating a token and signing it with a private key. Perhaps split those in 2 separate functions to make it more readable or just call it something like `createSignedToken`

controllers/integration/integration_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	issuedAt := time.Now().Unix()
+
+	token := jwt.NewWithClaims(jwt.SigningMethodES384, jwt.MapClaims{
+		"sub": sa.Mrn,
+		"iss": "mondoo/ams", // needs to be set to "mondoo/ams" to be accepted by Mondoo API
```
imilchev, line 166: I think the issuer should be a const as well

controllers/integration/integration_controller.go
```diff
@@ -0,0 +1,181 @@
...
+			Name:      mondoo.Spec.MondooCredsSecretRef.Name,
+			Namespace: mondoo.Namespace,
+		},
+	}
+	if err := r.Client.Get(context.Background(), client.ObjectKeyFromObject(mondooCreds), mondooCreds); err != nil {
+		r.log.Error(err, "failed to read Mondoo creds from secret")
+		return err
+	}
+
+	integrationMrn, ok := mondooCreds.Data[constants.MondooCredsSecretIntegrationMRNKey]
```
imilchev, line 110: I guess this is only set if we registered the operator with the conversion from short- to long-lived token.  If I were to create my own `MondooCredsSecret` that would not work. If that is true, then perhaps make the error message below a little more clear to say that it is not possible to enable console integration for a "normal" `MondooCredsSecret`.

imilchev: This also gets me thinking... Will there be any case in which you are using the flow that converts the tokens and don't want to have the console integration enabled?

joelddiaz: Not only will the 'integratiomrn' field only be set when doing the conversion from token-to-serviceaccount, it will only be set when doing a token-to-serviceaccount when the MondooAuditConfig.Spec.ConsoleIntegration was enabled. And the token better have had the `owner` claim in it because that is how the operator "saves" the integration MRN.

I suppose if you were clever enough, you could just create an Integration, manually exchange that token for a service account, and then save the Secret with both the `config` and the `integrationmrn` set in the Secret. Then you could set the MondooAuditConfig.Spec.ConsoleIntegration to enabled, and the mondoo-operator would never know that you did all the work outside of the operator...it would just work as if the operator had done the token exchange and the Secret generation on its own.

You can still bring your own token and the mondoo-operator will do the token-to-serviceaccount exchange. As long as MondooAuditConfig.Spec.ConsoleIntegration is not enabled, then mondoo-operator will do an /AgentManager/ExchangeRegistrationToken() API call instead of the /IntegrationManager/Register() API call. The ExchangeRegistrationToken() call will just do a "regular" token-to-serviceaccount exchange, with no association between the Mondoo service account and any integration.

imilchev (approved):
:shipit:

PR #336, opened 2022-05-12
🐛 Merge workflows for Helm and OLM to Publish

imilchev:


imilchev:
Merged OLM and Helm jobs into the Publish workflow. It seems people were complaining about not being able to use the secrets in re-usable workflows. See [here](https://github.com/actions/runner/issues/1413). Interestingly enough they fixed it but now it seems to be broken again...

imilchev:
Adding this for reference: https://github.com/actions/runner/issues/1490

joelddiaz (approved):

PR #341, opened 2022-05-13
set a condition if issues during integration CheckIn()

joelddiaz:
Condition will only appear if it was Degraded once. After clearing up, the condition will exist in the list, but be set to Status == false (with a "all is well" message).


imilchev (approved):

joelddiaz:
@imilchev FYI, I ended up rebasing this on top of #344 so that should go in before this PR.

imilchev:
Can this be merged?

Issue #343, opened 2022-05-14
mondoo operator waits too long to call checkin after create


vjeffrey:
**Describe the bug**
mondoo operator waits too long to call checkin after first creating 

**To Reproduce**
tested with minikube

- i set up the integration following the exact info i got from the ui
- saw that the token was exchanged for a service account right away, things reported in
- but there was no checkin update
- waited 10 minutes, then there was a checkin

the mondoo-operator needs to checkin right away after exchanging the service account for the flow to work well

imilchev:
should be fixed in #344

PR #344, opened 2022-05-14
attempt an immediate CheckIn() after Register()

joelddiaz:
When trading a token for a Mondoo service account, don't wait for the periodic integration controller to run and perform the first CheckIn(). Just make an attempt right after saving the Secret with the Mondoo service account.

Update test cases to expect the additional CheckIn() API call, and restructure the Integration controller to more easily allow perfoming a CheckIn() using only the Integration MRN and the Mondoo service account data.

Addresses #343 

### DIFF
--- a/controllers/integration/integration_controller.go
+++ b/controllers/integration/integration_controller.go
@@ -132,28 +132,6 @@ func (r *IntegrationReconciler) processMondooAuditConfig(mondoo v1alpha2.MondooA
 		return err
 	}
 
-	// serviceAccount := &mondooclient.ServiceAccountCredentials{}
-	// if err := json.Unmarshal(mondooCreds.Data[constants.MondooCredsSecretServiceAccountKey], serviceAccount); err != nil {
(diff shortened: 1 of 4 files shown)


joelddiaz:
cc @vjeffrey

controllers/integration/integration_controller.go
```diff
@@ -114,27 +114,62 @@ func (r *IntegrationReconciler) processMondooAuditConfig(mondoo v1alpha2.MondooA
...
 	}
+
+	serviceAccount := mondooCreds.Data[constants.MondooCredsSecretServiceAccountKey]
+
+	if err := r.IntegrationCheckIn(integrationMrn, serviceAccount); err != nil {
+		r.Log.Error(err, "failed to CheckIn() for integration", "integrationMRN", string(integrationMrn))
+		return err
+	}
+
+	// serviceAccount := &mondooclient.ServiceAccountCredentials{}
```
imilchev, line 135: i guess this can go away if it isn't needed

vjeffrey (approved):
lgtm!

PR #350, opened 2022-05-16
mondoo 0.2.13

joelddiaz:

### DIFF
--- a/charts/mondoo-operator/templates/deployment.yaml
+++ b/charts/mondoo-operator/templates/deployment.yaml
@@ -40,9 +40,6 @@ spec:
         - --leader-elect
         command:
         - /manager
-        env:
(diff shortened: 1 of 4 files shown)


charts/mondoo-operator/templates/deployment.yaml
```diff
@@ -40,9 +40,6 @@ spec:
         - --leader-elect
         command:
         - /manager
-        env:
```
imilchev, line 43: btw this was apparently added in the newer versions of `helmify`. That's why it ended up being committed. I guess you just have an older one.

imilchev (approved):

charts/mondoo-operator/templates/deployment.yaml
joelddiaz: I deleted my local helmify, and pulled in a new one to bring back those changes.

PR #356, opened 2022-05-20
🧹 set integration label when console integration is enabled

### LINKED ISSUE #354: add integration mrn label to all assets discovered with operator
looking forward to support the integrations screen that was in our internal demos today, we should ensure all k8s assets are labelled with `mondoo.com/integration-mrn` and the integration mrn they are attached to

this will allow us to filter the assets list to the assets that are part of the integration mrn/allow us to display the numbers needed for those

joelddiaz:
To follow allong with other integrations, set the 'mondoo.com/integration-mrn' label for scanned resources.

Update the webhook to accept this optional parameter, and set the label if the integration MRN was provided.

Update test cases.

Closes #354 


vjeffrey (approved):
thank you!

imilchev (approved):

PR #358, opened 2022-05-23
add long(er) lived container image cache

### LINKED ISSUE #304: Long-lived container image digest cache
We now have a cache to make sure we resolve container images just once per reconcile cycle (see #301). However, we still perform quite a lot of cycles since we monitor pods. We should consider having a long-lived cache that make sure we don't resolve the images once per reconcile cycle but maybe once per X minutes

joelddiaz:
Extend image caching beyond a single MondooAuditConfig Reconcile() loop.

Introduce an image cache that will keep an image -> imageDigest translation for up to 1 day before attempting to refresh.

Update test cases as needed, and write new test cases to excercise the cache.

Closes #304 


pkg/utils/mondoo/container_image_resolver.go
```diff
@@ -38,8 +39,6 @@ const (
 // "latest".
 var MondooOperatorTag = version.Version
 
-type getRemoteImageFunc func(ref name.Reference, options ...remote.Option) (*remote.Descriptor, error)
-
 type ContainerImageResolver interface {
```
imilchev, line 42: It seems like this struct no longer resolves anything. All the resolve logic is moved to the `ImageCache`. I guess it might even be better to just combine the 2 structs into 1.

pkg/imagecache/imagecache.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+type imageData struct {
+	url         string
+	lastUpdated time.Time
+}
+
+func (i *imageCache) GetImage(image string) (string, error) {
+
+	defer i.cleanup()
```
imilchev, line 33: If you call it with `defer` wouldn't that mean that it will only execute the cleanup after it returns the old image? I would expect that the cache would first check if the entry should be invalidated, potentially get the new image and then give me the result instead of returning an old entry

pkg/imagecache/imagecache.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+type ImageCacher interface {
+	GetImage(string) (string, error)
+}
+
+type imageCache struct {
+	images      map[string]imageData
+	imagesMutex sync.RWMutex
+	lastCleanup time.Time
```
imilchev, line 22: I think we can further simplify this and make it a little more "lazy". Instead of using this `lastCleanup` we can just use the `lastUpdated` of `imageData`. I would propose to do the following when `GetImage` is called:
1. Find the image in the cache
2. If the image is old, refresh it and add to cache
3. Return image

This would mean that we only check the image's TTL if it is actually being requested instead of looping the entire cache every time `GetImage` is called. Would also make the code simpler I believe

imilchev:
This would be a huge improvement as I already ran into a situation where the operator just exhausted the image pulls from Dockerhub and then this happens:

```
1.6535629212963006e+09 ERROR controller.mondooauditconfig Failed to declare nodes {"reconciler group": "k8s.mondoo.com", "reconciler kind": "MondooAuditConfig", "name": "mondoo-client", "namespace": "mondoo-operator", "error": "GET https://index.docker.io/v2/mondoo/client/manifests/5.39.0: TOOMANYREQUESTS: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"}
```

pkg/utils/mondoo/container_image_resolver.go
joelddiaz: I suppose the interface is still responsible for simply resolving the Mondoo client and the Mondoo operator image. The container image caching is more generic and can cache any images.

pkg/imagecache/imagecache.go
joelddiaz: With this approach, how do you propose cleaning up images that are not actively being queried?

We already do number 1 and number 2 and number 3.

The lastCleanup time is just so we periodically purge any images that haven't been queried/updated in a while. Perhaps the cleanup should run less often than the refreshPeriod to make it clear that it is meant to cleanup orphaned images.

imilchev: Yeah, I noticed but it seems overkill to me that we clean that every time an image is requested. I would even argue that not cleaning it all is also fine because even if you have it running for months and have a few hundred images cached that would be a few kilobytes. We store almost no data.

joelddiaz: We're not cleaning on every image request, just checking whether it is time to clean. Since there is no background process to do periodic cleanup, it's just using the invocation of GetImage() as the opportunity to do some periodic cleanup.

Either way, like you say, it's a tiny amount of data. I'll remove the cleanup piece and just put a comment about adding some kind of cleanup for someone that may try to use this image caching for other purposes (like maybe the registry scanning that we keep talking about).

PR #360, opened 2022-05-24
🧹 add Integration MRN for workload scanning

### LINKED ISSUE #359: set k8s labels in the workload scanning configuration
When a MondooAuditConfig has the consoleIntegration set to enabled, ensure that the workload scanning configuration file ( https://github.com/mondoohq/mondoo-operator/blob/main/controllers/inventory-deploy.yaml ) passed to the Mondoo client includes the console Integration MRN label ('mondoo.com/integration-mrn' set to the MRN as is done in the webhook https://github.com/mondoohq/mondoo-operator/blob/main/pkg/webhooks/handler/webhook.go#L159-L161 ).

joelddiaz:
Define a MondooInventory type to allow programatically generating the Inventory passed to the Mondoo client for workload scanning.

If/when the console integration is enabled on the MondooAuditConfig, set the Integration MRN just like we do for the admission controller.

Closes #359 


chris-rock (approved):
Thank you @joelddiaz

PR #367, opened 2022-05-30
🧹 Revised resource requests/limits for Mondoo client

imilchev:
Tested the Mondoo client container by doing different scans and measuring the resource utilisation using `docker stats`.

I did workloads scan for a 3-node Kubernetes cluster and the average memory usage was around 150-160MB. The CPU is shown in % so it was not very useful. I suppose it is showing % per core or something as it spiked to 150%+. However, CPU limits should be fine as it will just throttle the scan but will not kill the pod as opposed to a lowed memory limit.

I scanned a few Kubernetes manifests (what the scan API is doing) and the memory usage was about 50MB and a very little CPU utilisation of about 10-15% for a very short period.


controllers/scanapi/resources.go
```diff
@@ -73,10 +73,20 @@ func ScanApiDeployment(ns, image string, m v1alpha2.MondooAuditConfig) *appsv1.D
...
-						Image:     image,
-						Name:      "mondoo-client",
-						Command:   []string{"mondoo", "serve", "--api", "--config", "/etc/opt/mondoo/config/mondoo.yml", "--token-file-path", "/etc/opt/mondoo/token/token"},
-						Resources: k8s.DefaultMondooClientResources,
+						Image:   image,
+						Name:    "mondoo-client",
+						Command: []string{"mondoo", "serve", "--api", "--config", "/etc/opt/mondoo/config/mondoo.yml", "--token-file-path", "/etc/opt/mondoo/token/token"},
+						Resources: corev1.ResourceRequirements{
+							Limits: corev1.ResourceList{
+								corev1.ResourceMemory: resource.MustParse("300M"),
```
imilchev, line 81: there seems to be some initial spike in the memory. If the limit is too low the pod just gets killed...

joelddiaz: Did you try running this through a memory profiler to get the exact peak memory usage? I worry that these default numbers may work for passing integration testing, but could become problematic on environments with more k8s resources...

imilchev: The limit is sufficient. The operational memory for the API is about 50-80MB. However, initially it spikes for some reason. I noticed that if the memory limit is below 200 the pod just gets killed right after it is created.

joelddiaz: I suppose we can go with these limits as long as someone can get around them without needing a new version of the mondoo-operator. Shouldn't the Scan API Deployment definition also be checking against the contents of .spec.scanner.resources?

imilchev: Applying the requirements from .spec.scanner.resources wasn't implemented because now we have multiple modes for the Mondoo client running at the same time and each of them has different requirements. I say we leave it as is for now and properly fix it once we make sure every scan goes via the scan API.

joelddiaz: I worry that we are bringing down the requests/limits which could have unintended consequences. I'd rather avoid having someone opening an issue after upgrading, and then our solution is "wait for an updated release of mondoo-operator". What is the harm in having the scanner resource/limits fields affect the Scan API Deployment?

imilchev: None, it's just that the requirements will be slightly higher for the scan API than it actually needs them to be. I will change the code

joelddiaz (approved):

PR #368, opened 2022-05-30
🧹 Remove v1alpha1

imilchev:
- [x] Updated `controller-gen` to `0.9.0`
- [x] Removed `v1alpha1`
- [x] Fixed `manifests` Makefile target

The path that I removed from the `manifests` target was causing it to crash. The older version of `controller-gen` actually yielded no errors and just silently did nothing. After I updated I saw that there was an error and that's why the manifests were not being updated.

Since we completely broke people's clusters when releasing `v1alpha2` and that required manually deleting the CRDs with the older API version, we can now directly delete the old version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,7 +171,7 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 .PHONY: controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.0)
+	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0)
 
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 .PHONY: kustomize
--- a/config/crd/bases/k8s.mondoo.com_mondooauditconfigs.yaml
+++ b/config/crd/bases/k8s.mondoo.com_mondooauditconfigs.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -171,7 +171,7 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 .PHONY: controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0)
+	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.0)
```
joelddiaz, line 174: This separates us from operator sdk/framework. What problem is this fixing?

imilchev: Are they pinned on `v0.8.0`? I realized that I was running `v0.7.0` so it might be the cases that `v0.8.0` would have fixed the issue as well. The problem was that the command was not okay and `v0.7.0` silently did nothing instead of returning an error...

joelddiaz: Yes, doing an `operator-sdk init` leaves you with a Makefile that will will resolve controller-gen to 0.8.0.

```
projects/mytest  % operator-sdk init --domain exmaple.com --repo github.com/example/memcached-operator
Writing kustomize manifests for you to edit...
Writing scaffold for you to edit...
Get controller runtime:
$ go get sigs.k8s.io/controller-runtime@v0.11.2
Update dependencies:
$ go mod tidy
Next: define a resource with:
$ operator-sdk create api
projects/mytest  % ack CONTROLLER_TOOLS_VERSION Makefile
CONTROLLER_TOOLS_VERSION ?= v0.8.0
        GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
```

imilchev: yes, `v0.8.0` works as well. I reverted that change

joelddiaz (approved):

PR #370, opened 2022-05-31
⭐ Document versioning and upgrade strategy

### LINKED ISSUE #365: Document release strategy
We should document our release strategy and make it transparent for end-user when and how they are expected to upgrade the operator. My proposal would be to have an upgrade path that should go through every minor release of the operator. Documenting this would allow us to also have a clear plan when and how we can remove cleanup code and backwards compatibility hacks

imilchev:
Closes #365 

### DIFF
--- a/docs/operator-upgrades.md
+++ b/docs/operator-upgrades.md
@@ -3,7 +3,7 @@ The Mondoo operator versions are expressed as `x.y.z`, where `x` is the major ve
 
 When upgrading the operator it is important to note that skipping patch versions is possible but skipping minor versions is not. Upgrading from `v0.2.0` directly to `v0.4.0` is possible but can result in the operator not functioning as expected and/or unused operator resource being left behind in the cluster. Performing such an upgrade will require manual actions to ensure the operator is fully functional.
 
-Never upgrade the operator by simply changing the tag for the Mondoo operator container image!
+**Never upgrade the operator by simply changing the tag for the Mondoo operator container image!**
 
 ## Recommended operator upgrade approach
 Follow the steps below to ensure smooth Mondoo operator upgrade procedure:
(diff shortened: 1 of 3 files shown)


imilchev:
@joelddiaz please check whether what I wrote makes sense. If you agree, I will ask Letha to check it textually

docs/operator-upgrades.md
```diff
@@ -0,0 +1,39 @@
+# Upgrading the operator
+The Mondoo operator versions are expressed as `x.y.z`, where `x` is the major version, `y` is the minor version, and `z` is the patch version, following [Semantic Versioning](https://semver.org/) terminology. The operator releases are done in a way that makes sure there are no breaking changes between 2 adjacent minor versions. For example, upgrading the operator from `v0.2.15` to `v0.3.0` is possible without any manual actions. The operator will automatically execute any migration and/or cleanup steps needed.
+
+When upgrading the operator it is important to note that skipping patch versions is possible but skipping minor versions is not. Upgrading from `v0.2.0` directly to `v0.4.0` is possible but can result in the operator not functioning as expected and/or unused operator resource being left behind in the cluster. Performing such an upgrade will require manual actions to ensure the operator is fully functional.
```
joelddiaz, line 4: Just to confirm my understanding: the text below makes it clear that upgrading between minor releases that are the '.0' patch is okay (eg upgrade from 0.2.5 to 0.3.3 is valid). And this means that any transitional cleanup code would be introduced at the '0.x.0' patch and be removed on the next '0.y.0' patch. Correct?

imilchev: That was my idea at least. In this way we don't make it too complicated for the users

imilchev:
> The other overarching question I have regarding this is whether this documented behavior is intended while we are still in major version '0', but will it change to a "must upgrade to every major release" (so users can skip minor releases) once the major version hits '1'?

I think eventually we are going to go there but I guess we need to define some process for the time being and then we can have another look at it once we are more stable

joelddiaz (approved):

PR #371, opened 2022-06-01
mondoo 0.2.16

joelddiaz:


imilchev:
If we agree to follow what I described in https://github.com/mondoohq/mondoo-operator/pull/370 this should be `v0.3.0` since it removes the `v1alpha1`. Another benefit of having it as `v0.3.0` is that after we release it we can get rid of all kind the cleanup code that we currently have and have a fresh start for our new versioning policy.

PR #373, opened 2022-06-01
mark some of the cleanup functions that we can remove with 0.4.0

joelddiaz:


joelddiaz:
This doesn't really need to wait on #372 , but logically this would follow the 0.3.0 release as we mark the things that can be removed as part of the 0.4.0 release.

PR #374, opened 2022-06-01
🐛 Fix helm release workflow

imilchev:
The issue with the helm release workflow was that the job that execute the integration tests was also doing the actual release of the chart. The workflow tried to release the same chart 3 times (for each k8s version tested).

This PR splits off the actual release of the chart to a separate job that runs only after all integration tests have passed.


imilchev:
I suggest we also release `v0.4.1` after this PR merges just to verify that the workflow works correctly this time

PR #376, opened 2022-06-02
0.3.1

imilchev:

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=mondoo-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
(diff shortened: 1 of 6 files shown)


bundle.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=mondoo-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.19.0+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.21.0
```
joelddiaz, line 9: we haven't yet made the jump to 1.21

joelddiaz (approved):
It doesn't look significant, but there is that odd label about operator-sdk v1.21...

bundle.Dockerfile
imilchev: I think my local `operator-sdk` got updatged to 1.21 and that's why this happened...

PR #377, opened 2022-06-02
Add check and tests for missing Service Account

### LINKED ISSUE #298: Enforce the need for a ServiceAccount name when webhook enabled in non-default namespace
The ScanAPI Deployment will need to make k8s API calls. While there is a ready-made ServiceAccount for precisely this purpose in the 'mondoo-operator' Namespace (for the Workload scanning), ~~we need to ensure (for now) that the .spec.Workloads.ServiceAccount field is filled out when setting up the ScanAPI Deployment in a Namespace that isn't 'mondoo-operator'.~~ , we need to ensure that the ServiceAccountName specified in .spec.Scanner.ServiceAccountName exists before creating the Depoyment.

This would be handled with something like:
```
if Mondoo.Namespace != "mondoo-operator" {
  exists := doesServiceAccountExists(Mondoo)
  if !exists {
    // log error
   // set condition
   return
  }
}
```

Criteria:
Write up some test case(s) to assert this behavior (should be failing if the tests are written first).

A functioning setup would mean that when a MondooAuditConfig in a non-default (ie non `mondoo-operator`) Namespace would log the error and return an overall error in the mondooauditconfig controller's Reconcile() loop and keep from starting up the ScanAPI Deployment/Pod.

czunker:
These tests check for missing Service Accounts when creating a Deployment.

WIP because it works, but the resource cleanup is missing.

Fixes #298


joelddiaz:
I see you're plugging into the live/e2e integration environment (which is totally fine). I hadn't even thought of that, and just assumed we could get away with a more simple unit test in https://github.com/mondoohq/mondoo-operator/blob/main/controllers/scanapi/deploy_test.go

Either way is fine, I just wanted to communicate that there were options.

czunker:
Thanks for the hint. I currently have both. I'm not sure how this will work out.

czunker:
I figured out, why the resource cleanup didn't work as expected: https://github.com/kubernetes-sigs/controller-runtime/issues/880

It is a limitation of `envtest`.

controllers/scanapi/deploy.go
```diff
@@ -35,6 +36,16 @@ var logger = ctrl.Log.WithName("scan-api-deploy")
 // Deploy deploys the scan API for a given MondooAuditConfig. The function checks if the scan API is already deployed.
 // If that is the case, the existing resources are compared with the ones that are desired and the necessary updates are applied.
 func Deploy(ctx context.Context, kubeClient client.Client, ns, image string, mondoo mondoov1alpha2.MondooAuditConfig) error {
+	//Check whether specified SA exists in the namespace, just in case, e.g. when Deployment is spread across namespaces
```
imilchev, line 39: it looks like you have the same check in `controllers/workloads.go`. Perhaps extract it somewhere and just re-use it

czunker:
I guess, this is the reason, why the integration tests are failing:
```
E0603 12:41:02.097969       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.23.4/tools/cache/reflector.go:167: Failed to watch *v1.ServiceAccount: failed to list *v1.ServiceAccount: serviceaccounts is forbidden: User "system:serviceaccount:mondoo-operator:mondoo-operator-controller-manager" cannot list resource "serviceaccounts" in API group "" at the cluster scope
```

joelddiaz:
> I guess, this is the reason, why the integration tests are failing:
> 
> ```
> E0603 12:41:02.097969       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.23.4/tools/cache/reflector.go:167: Failed to watch *v1.ServiceAccount: failed to list *v1.ServiceAccount: serviceaccounts is forbidden: User "system:serviceaccount:mondoo-operator:mondoo-operator-controller-manager" cannot list resource "serviceaccounts" in API group "" at the cluster scope
> ```

Ah yeah, interacting with ServiceAccount resources is a new thing that we haven't given the mondoo-operator permissions for previously.

czunker:
As it turned out, this change has security implications, because the operator would be able to read all Kubernetes Service Accounts. Think of it like reading `/etc/passwd` on Linux. With GKE Workload Identity or AWS IAM role association, this might even have more impact.

We will try another approach with #382

Issue #378, opened 2022-06-03
Migrate Kubernetes resources scanning to a CronJob


imilchev:
Do not use a `Deployment` running `mondoo serve` for scanning the Kubernetes resources. Instead use the kubernetes service account from the scan API instance and define a `CronJob` to invoke the scan API and trigger the scan periodically.

Will require creating a new binary for communicating with the scan API for this purpose.

PR #383, opened 2022-06-07
🧹 Merge operator and webhook binary into a single CLI binary

imilchev:
We have 2 separate binaries now that are build separately - 1 for the operator and 1 for the validating webhook. To implement #378 another binary is going to be needed. Having 3 binaries would make things even more complicated. Since we have just 1 container with both binaries in it, it would make sense to just merge them into a single thing.

To simplify the build process this PR creates 1 CLI using `cobra`. That would allow us to easily extend the operator functionality by just adding commands.


pkg/webhooks/handler/webhook.go
```diff
@@ -72,7 +72,8 @@ func NewWebhookValidator(client client.Client, mode, scanURL, token, integration
 			ApiEndpoint: scanURL,
 			Token:       token,
 		}),
-		clusterID: clusterID,
+		integrationMRN: integrationMRN,
```
imilchev, line 75: This was a bug that I found, which was introduced in #354. The `integrationMRN` was never set, so it was never actually used

config/manager/kustomization.yaml
```diff
@@ -14,4 +14,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: ghcr.io/mondoohq/mondoo-operator
-  newTag: v0.3.1
+  newTag: sha256-f27f50752316d951db39763ba380a3f2de9c8168.sig
```
joelddiaz, line 17: Normally, we leave this the same way the `release.sh` script modified it.

config/webhook/kustomization.yaml
```diff
@@ -7,4 +7,4 @@ resources:
 images:
 - name: controller
   newName: ghcr.io/mondoohq/mondoo-operator
-  newTag: v0.0.1
+  newTag: sha256-f27f50752316d951db39763ba380a3f2de9c8168.sig
```
joelddiaz, line 10: Same here.

config/manager/kustomization.yaml
imilchev: I overlooked those... will revert them 🥔

PR #384, opened 2022-06-07
⭐ Kubernetes resource scanning CronJob

### LINKED ISSUE #378: Migrate Kubernetes resources scanning to a CronJob

imilchev:
Note that this PR is based on #383.

- [x] Extend Mondoo scan API client to be able to send scan jobs
- [x] Create a new command to trigger the Kubernetes resource scanning
- [x] Use the resources scanning service account for the scan API deployment
- [x] Deploy the new binary as a `CronJob` that is triggered hourly
- [x] Add cleanup code for old Kubernetes resource scanning
- [x] Extract deployment of Scan API outside of `admission` and `k8s_scan` since the current approach results in endless loop
- [x] Adjust tests 
- [x] Verify left TODOs

When this is merged, we should release it as `v0.4.0`

Closes #378 


tests/framework/utils/rand.go
```diff
@@ -0,0 +1,29 @@
+/*
```
czunker, line 1: As RandString is also used in the controller packages, perhaps this should be moved to a general utils package?

czunker:
As this changes the architecture, could we also include an updated version of https://github.com/mondoohq/mondoo-operator/blob/main/docs/img/architecture.svg?

tests/framework/utils/rand.go
imilchev: It's used only in the test files. That's why it is under the test utils.

controllers/k8s_scan/workloads_test.go
```diff
@@ -0,0 +1,136 @@
+/*
```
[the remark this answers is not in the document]
imilchev: I think it ended up being committed by accident. It should be removed, indeed.

imilchev:
> As this changes the architecture, could we also include an updated version of https://github.com/mondoohq/mondoo-operator/blob/main/docs/img/architecture.svg?

Updated the graph

controllers/k8s_scan/resources.go
```diff
@@ -0,0 +1,120 @@
...
+							Containers: []corev1.Container{
+								{
+									Image:           image,
+									ImagePullPolicy: corev1.PullIfNotPresent,
+									Name:            "mondoo-client",
+									Command:         []string{"/mondoo-operator"},
+									Args: []string{
+										"k8s-scan",
+										"--scan-api-url", scanApiUrl,
+										"--token-file-path", "/etc/webhook/token",
```
joelddiaz: seeing the word 'webhook' here surprised me a bit. Maybe /etc/scanapi/token ?

pkg/mondooclient/client.go
```diff
@@ -28,7 +29,7 @@ type Client interface {
 
 	HealthCheck(context.Context, *HealthCheckRequest) (*HealthCheckResponse, error)
 	RunKubernetesManifest(context.Context, *KubernetesManifestJob) (*ScanResult, error)
-	ScanKubernetesResources(context.Context) (*ScanResult, error)
+	ScanKubernetesResources(ctx context.Context, integrationMrn string) (*ScanResult, error)
```
joelddiaz, line 32: There doesn't seem to be any benefit to defining the variable names `ctx` and `integrationMrn` here in the interface definition.

pkg/utils/k8s/integration_mrn.go
```diff
@@ -0,0 +1,55 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetIntegrationMrnForAuditConfig retrieves the integration-mrn for a MondooAuditConfig. This is only relevant if ConsoleIntegration
+// is enabled. When ConsoleIntegration is disabled an empty string is returned.
+func GetIntegrationMrnForAuditConfig(ctx context.Context, kubeClient client.Client, auditConfig v1alpha2.MondooAuditConfig) (string, error) {
+	if !auditConfig.Spec.ConsoleIntegration.Enable {
+		// sending an empty integrationMRN means the webhook will run w/o setting integration
```
joelddiaz, line 34: "means the webhook or k8s resource scan will run"

config/manager/kustomization.yaml
```diff
@@ -14,4 +14,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: ghcr.io/mondoohq/mondoo-operator
-  newTag: v0.3.1
+  newTag: sha256-f27f50752316d951db39763ba380a3f2de9c8168.sig
```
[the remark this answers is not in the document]
imilchev: ah... these files... we have to find a way to address this...

cmd/mondoo-operator/k8s_scan/cmd.go
```diff
@@ -0,0 +1,62 @@
...
+		res, err := client.ScanKubernetesResources(context.Background())
+		if err != nil {
+			logger.Error(err, "failed to trigger a Kubernetes resources scan")
+		}
+
+		// TODO: print some more useful info
+		if res.Ok {
+			logger.Info("Kubernetes resources scan successful", "worst score", res.WorstScore.Value)
+		} else {
+			logger.Error(fmt.Errorf("scan API returned not OK. %v", res.WorstScore), "Kubernetes resources scan was not successful")
```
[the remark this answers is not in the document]
imilchev: yes, that was the initial plan but it seems I forgot to add that bit of code

cmd/mondoo-operator/k8s_scan/cmd.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+		token := strings.TrimSuffix(string(tokenBytes), "\n")
+
+		client := mondooclient.NewClient(mondooclient.ClientOptions{
+			ApiEndpoint: *scanApiUrl,
+			Token:       token,
+		})
+
+		logger.Info("triggering Kubernetes resources scan")
+		res, err := client.ScanKubernetesResources(context.Background())
```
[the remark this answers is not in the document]
imilchev: the HTTP client in the `mondooclient` already has a 30s timeout. I guess that should be sufficient

pkg/utils/k8s/resources_requirements.go
```diff
@@ -8,13 +8,13 @@ import (
...
-		corev1.ResourceMemory: resource.MustParse("500M"),
-		corev1.ResourceCPU:    resource.MustParse("400m"),
+		corev1.ResourceMemory: resource.MustParse("300M"),
+		corev1.ResourceCPU:    resource.MustParse("1"),
 	},
 
 	Requests: corev1.ResourceList{
 		corev1.ResourceMemory: resource.MustParse("180M"),
-		corev1.ResourceCPU:    resource.MustParse("150m"),
+		corev1.ResourceCPU:    resource.MustParse("400m"),
```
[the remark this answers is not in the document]
imilchev: well, we know that 1 core gives the the fastest performance. If we set the requests to `150m` it is possible that the pod gets scheduled on a node that is over-provisioned. This means that practically the limit for the pod is `150m` which is not sufficient for the operation of the scan API.

In a production setup you probably even want the pod to request 1 CPU.

controllers/k8s_scan/resources.go
imilchev: indeed, that should be the case also for the admission controller tbh. I will change it there as well.

pkg/mondooclient/client.go
imilchev: I like adding the names when there are default types parameters such as `string`, `int`, etc. The benefit is that the name of the parameter acts as a documentation when you look at the interface. For example, right now, you read this and you know that the second parameter is supposed to be the `integrationMrn`. If it just stated `string`, then you have no idea what exactly that string is supposed to be. The thing is that if you want to name 1 parameter in the declaration of a function, you need to name them all...

docs/img/architecture.drawio
```diff
@@ -1 +1 @@
...
...5ZfLZ+T4HhNEtcVEypq6Zs9jHHC020EnuxiIEhM3nYSBaQFyuKeGSU5C4PyDGsQB3GmCPCCOnm9xLzlCzLGlK+qJdySSXeY1oiDjbCxd518ohlzFnjGQ7Lhg0LGnzS+wZY2mEMmq8w9gFsOJCYvsbOJsKzgwKnMMBic4RYfM0wObwWQG2FIDfp4E83bqYzylZY69boLuGdFSF1AamAikwaiA1nwpRW0FUM4aBmHW2EhdecvFuu0KMIC4KtAAWsQcswBOP4jgCWZ67i9kPPQ5DsNzCUES3zMmGcCAhmHTMVPfkWPkIp8ixnoqc4WlysvifB1gse4EchloSUsNjv9kwx2r1ed5UmSgQIVeoCNmkjPvUowQGV4V1VgWx8LmlScVOofsPcb6X2MEtp1VgBX5s/0X2TxvLpHFh583LXfnm5V62WhOS5WXDsqUG4pB5iDf4jTO/BJNGehkKIMcPVeXVOVkgj7cid0RywLSI1bB4C1dC0laQh3JvdQRiiNVsuiF23YxkFOGfcJWOl2C9oZjwdEH2TLMvW7HRFHFKzhyEsJxUK2vNulzSL3RjbFXySQ7Umg059l2ytpILXa8jxBW6Do/weAZHCoFCRpMbulKLWW6JNpBUuB3+2CbCO8V/EMsiMRUuhLIQBllX6dO8VR1qajZHPu8ZaQ2rRr49bwEFxotWUPBSFXTcsoLaXVfQtOuUMbgvOcjqpKZ0rvrBUdxY4Ij5bMROkz3Xwa+h0agEexYa5sR+htBQX/rERu6e0z6eR/efb+QD/WI0mYwrLAz6tJXHm+2H+NPN8l/mLUJOv35b6u8GL5vcpXxelu/1JrlB54W/7TZdy5b6SeAjw54ncsfQv4+j9DmqEqmfqdgUip1koglss0ep1xSDTZ8QSpK6+s2gZx8EDPuRgtaYPJWgVVMl+x4jbIs0L/Tp3dtugez6LcE+3tP1lqgOO3hNqE1RoAD2t31n+cHnXxb34G5PDfvKnN3fXYO3+bLPRAoe5aJ5fCx1wr1j4VgbR8Zv6MbX6OrVO+ifRVfXb6y10WW9xtEZaVr1JEXuzyEkeI0i9fijvxI2fx/uQsKCkTWq5I4smr0QtE0BePpMTDjdC7ZcsY7kA67+Ga18Sr+3Ox6DboijCFOSocoZDYLkBahnkvj4jAy01G6POLEUzeJfLBmZxX+BzKtf</diagram></mxfile>
```
joelddiaz, line 1: This diagram mentions a "global scan api". That isn't really part of this PR.

pkg/utils/k8s/resources_requirements.go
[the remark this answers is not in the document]
joelddiaz: If the concern is overprovisioning, then we're still exposed here (of course to a lesser extent as we're trying to guarantee ~ 2.5x more minimum CPU). But I suppose this is easy enough to adjust/tune as we gather more real world data.
And we're reducing the steady-state CPU requirements by dropping the long-lived k8s resource scanning Pod, so its probably close to a wash - cluster resource requirements-wise - with this PR. Lets see what happens...

docs/img/architecture.drawio
imilchev: I renamed the service but generally the graph remains the exactly the same. Just the naming was confusing

joelddiaz (approved):

PR #386, opened 2022-06-14
v0.4.0

imilchev:
also deletes cleanup code


pkg/utils/k8s/resources_requirements.go
```diff
@@ -24,7 +24,7 @@ import (
 // defaultMondooClientResources for Mondoo Client container
 var DefaultMondooClientResources corev1.ResourceRequirements = corev1.ResourceRequirements{
 	Limits: corev1.ResourceList{
-		corev1.ResourceMemory: resource.MustParse("300M"),
```
imilchev, line 27: I forgot that when the scan API starts it actually uses increased memory at the very beginning and then it goes down. I remember it shot a bit over 300MB which means that it can potentially get killed. Therefore increasing the memory limit should fix the failing tests

joelddiaz (approved):

Issue #387, opened 2022-06-16
Add integration label for node scanning


joelddiaz:
When a MondooAuditConfig has consoleIntegration enabled, the node scanning Pod definition should include the integration MRN as a label so that those assets are associated with the IntegrationMRN.

For node scanning we pass an inventory.yaml to the Mondoo client, so fill out the .spec.assets[].labels to contain the integration MRN.

label should be "mondoo.com/integration-mrn": "//integrationMRNGoesHere"

PR #388, opened 2022-06-16
🧹 add integration MRN label for node scanning

joelddiaz:
When the MondooAuditConfig has consoleIntegration enabled, ensure
that we set up a Mondoo inventory that includes the inventory MRN
label like we already do for cluster scanning and the webhook.

closes: #387 


joelddiaz:
cc @vjeffrey

controllers/nodes/resources.go
```diff
@@ -173,9 +172,46 @@ func OldConfigMapName(prefix string) string {
 	return fmt.Sprintf("%s%s", prefix, OldInventoryConfigMapSuffix)
 }
 
-func Inventory(node corev1.Node, m v1alpha2.MondooAuditConfig) string {
-	inventory := string(inventoryYaml)
-	return strings.ReplaceAll(inventory, InventoryNodeNamePlaceholder, node.Name)
+func Inventory(node corev1.Node, integrationMRN string, m v1alpha2.MondooAuditConfig) (string, error) {
+	inv := inventory.MondooInventory{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "mondoo-k8s-inventory",
```
imilchev, line 178: can we change the name to `mondoo-node-inventory` or something along these lines?

imilchev (commented):
Apart from that nit it looks good! Let's ship :shipit:

imilchev (approved):

vjeffrey (approved):

controllers/nodes/resources.go
```diff
@@ -89,6 +89,10 @@ func CronJob(image string, node corev1.Node, m v1alpha2.MondooAuditConfig) *batc
 											Value: "on",
 										},
 									},
+									SecurityContext: &corev1.SecurityContext{
+										//RunAsUser:    pointer.Int64(9999),
```
imilchev, line 93: why the comment?

controllers/nodes/resources.go
```diff
@@ -89,6 +89,10 @@ func CronJob(image string, node corev1.Node, m v1alpha2.MondooAuditConfig) *batc
...
+										RunAsNonRoot: pointer.Bool(true),
```
imilchev, line 94: does that even work? The user within the mondoo client container is actually `root`. See [here](https://github.com/mondoohq/client/blob/master/Dockerfile#L27-L33)

controllers/nodes/resources.go
joelddiaz: That was some other thing I was working on...I'll remove

controllers/nodes/resources.go
joelddiaz: Same thing...working on a separate thing, and it snuck into this PR

PR #391, opened 2022-06-17
🐛 Make sure Deployments and CronJobs are updated when resource requests/limits change

imilchev:
I noticed that the operator doesn't compare the resource requests/limits for Deployments and CronJobs, so they never get updated. This PR fixes the issue


czunker (approved):
LGTM

PR #395, opened 2022-06-20
🧹 Add operator version to status of `MondooAuditConfig`

### LINKED ISSUE #380: Add ready check for the operator verifying that all MondooAuditConfigs have been reconciled
The goal is to have the Mondoo operator stating it is ready only if all `MondooAuditConfig` CRs have been reconciled at least once by the current version of the operator.

First, we need to extend the `MondooAuditConfig.Status` and add an extra property `ReconciledByOperatorVersion` (or some other name). That property should store the version of the Mondoo operator that last reconciled the CR.

After the status is extended, the operator should implement another ready check [here](https://github.com/mondoohq/mondoo-operator/blob/main/main.go#L123). The check should list all `MondooAuditConfig` and verify that the `ReconciledByOperatorVersion` field is equal to the current version of the operator. Only if that is the case, the operator should state it is ready.

czunker:
This adds the version of the operator which reconciled the `MondooAuditConfig`
to the `Status` section of the CR.

The new readiness check waits for all `MondooAuditConfigs` to include this version before the operator is considered ready.

Fixes #380

### DIFF
--- a/controllers/mondooauditconfig_controller.go
+++ b/controllers/mondooauditconfig_controller.go
@@ -271,11 +271,6 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	// Update status.ReconciledByOperatorVersion to the running operator version
 	if mondooAuditConfig.Status.ReconciledByOperatorVersion != version.Version {
 		mondooAuditConfig.Status.ReconciledByOperatorVersion = version.Version
-		err := r.Status().Update(ctx, mondooAuditConfig)
-		if err != nil {
-			log.Error(err, "Failed to update mondoo status")
-			return ctrl.Result{}, err
-		}
 	}
 
 	if err := mondoo.UpdateMondooAuditStatus(ctx, r.Client, mondooAuditConfigCopy, mondooAuditConfig, log); err != nil {
--- a/controllers/mondooauditconfig_controller.go
+++ b/controllers/mondooauditconfig_controller.go
@@ -54,6 +54,7 @@ import (
 	"go.mondoo.com/mondoo-operator/pkg/mondooclient"
 	"go.mondoo.com/mondoo-operator/pkg/utils/k8s"
 	"go.mondoo.com/mondoo-operator/pkg/utils/mondoo"
+	"go.mondoo.com/mondoo-operator/pkg/version"
 )
 
 const finalizerString = "k8s.mondoo.com/delete"
@@ -64,6 +65,7 @@ type MondooAuditConfigReconciler struct {
 	Scheme                 *runtime.Scheme
 	MondooClientBuilder    func(mondooclient.ClientOptions) mondooclient.Client
 	ContainerImageResolver mondoo.ContainerImageResolver
+	MondooAuditConfig      *v1alpha2.MondooAuditConfig
 }
 
 // Embed the Default Inventory for CronJob and Deployment Configurations
@@ -108,6 +110,7 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	mondooAuditConfig := &v1alpha2.MondooAuditConfig{}
 
 	err := r.Get(ctx, req.NamespacedName, mondooAuditConfig)
+	r.MondooAuditConfig = mondooAuditConfig
 
 	if err != nil {
 		if errors.IsNotFound(err) {
@@ -258,6 +261,16 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	if !statusPodNames.Equal(podListNames) {
 		mondooAuditConfig.Status.Pods = podListNames.List()
+		err := r.Status().Update(ctx, mondooAuditConfig)
+		if err != nil {
+			log.Error(err, "Failed to update mondoo status")
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Update status.ReconciledByOperatorVersion to the running operator version
+	if mondooAuditConfig.Status.ReconciledByOperatorVersion != version.Version {
+		mondooAuditConfig.Status.ReconciledByOperatorVersion = version.Version
 		err := r.Status().Update(ctx, mondooAuditConfig)
 		if err != nil {
 			log.Error(err, "Failed to update mondoo status")
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -2,7 +2,6 @@ package health
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 
@@ -17,8 +16,8 @@ type HealthChecks struct {
 	Log    logr.Logger
 }
 
-func (h *HealthChecks) CheckMondooAuditConfigsForOperatorVersion(req *http.Request) error {
-	ctx := context.TODO()
+func (h *HealthChecks) AreAllMondooAuditConfigsReconciled(req *http.Request) error {
+	ctx := context.Background()
 	mondooAuditConfigs := &v1alpha2.MondooAuditConfigList{}
 	if err := h.Client.List(ctx, mondooAuditConfigs); err != nil {
 		h.Log.Error(err, "error listing MondooAuditConfigs")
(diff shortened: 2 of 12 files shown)


imilchev:
> This includes changes to the `MondooAuditConfig` CRD. Do we bump the version or are there more changes that we should integrate?

This is just an update in the `Status` of the CRD which is considered an "internal" thing. Nothing really changes for the end-user. We can release this as a patch version just fine.

controllers/mondooauditconfig_controller.go
```diff
@@ -265,6 +268,16 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 	}
 
+	// Update status.ReconciledByOperatorVersion to the running operator version
+	if mondooAuditConfig.Status.ReconciledByOperatorVersion != version.Version {
+		mondooAuditConfig.Status.ReconciledByOperatorVersion = version.Version
+		err := r.Status().Update(ctx, mondooAuditConfig)
```
imilchev, line 274: You don't need to update here. The code on line 281 already performs the update if it's necessary. Check what that function does. Here you should just set the mondooAuditConfigCopy.Status.ReconciledByOperatorVersion = version.Version`

tests/integration/audit_config_base_suite.go
```diff
@@ -300,6 +301,32 @@ func (s *AuditConfigBaseSuite) testMondooAuditConfigAdmissionMissingSA(auditConf
...
+			SkipContainerResolution: true,
+		},
+	}
+	s.Require().NoErrorf(
+		s.testCluster.K8sHelper.Clientset.Create(s.ctx, operatorConfig), "Failed to create MondooOperatorConfig")
+
+	// Enable nothing
+	zap.S().Info("Create an audit config that enables nothing.")
+	s.NoErrorf(
+		s.testCluster.K8sHelper.Clientset.Create(s.ctx, &auditConfig),
```
imilchev, line 323: it's a little strange because the function is called `testMondooAuditConfigEmpty` but the actual `MondooAuditConfig` is provided as a parameter. It actually depends on the caller of this function whether the `MondooAuditConfig` will enable anything or not

tests/integration/audit_config_test.go
```diff
@@ -13,6 +13,11 @@ type AuditConfigSuite struct {
 	AuditConfigBaseSuite
 }
 
+func (s *AuditConfigSuite) TestReconcile_Empty() {
```
imilchev, line 16: Is there any gain in testing this? We could just add the version check in all the other tests I suppose...

controllers/mondooauditconfig_controller.go
czunker: Thanks. Fixed this.

tests/integration/audit_config_base_suite.go
czunker: I changed the wording to make more clear what's happening.

I think, that the test depends on the `MondooAuditConfig` of the caller is intentional. This is how the tests were designed to make it easier to call them in different contexts.

tests/integration/audit_config_test.go
czunker: I now opted for both.

I'd like to keep this test to spot bugs that are independent of any resources being created.

pkg/health/health.go
```diff
@@ -0,0 +1,40 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type HealthChecks struct {
+	Client client.Client
+	Log    logr.Logger
+}
+
+func (h *HealthChecks) CheckMondooAuditConfigsForOperatorVersion(req *http.Request) error {
+	ctx := context.TODO()
```
imilchev, line 21: use `context.Background()` instead of `TODO`. The `TODO` context shows up the same as TODOs left in the code.

pkg/health/health.go
```diff
@@ -0,0 +1,40 @@
...
+	"go.mondoo.com/mondoo-operator/pkg/version"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type HealthChecks struct {
+	Client client.Client
+	Log    logr.Logger
+}
+
+func (h *HealthChecks) CheckMondooAuditConfigsForOperatorVersion(req *http.Request) error {
```
imilchev, line 20: I think a name like `AreAllMondooAuditConfigsReconciled` is more suitable here

pkg/health/health.go
```diff
@@ -0,0 +1,40 @@
...
+	ctx := context.TODO()
+	mondooAuditConfigs := &v1alpha2.MondooAuditConfigList{}
+	if err := h.Client.List(ctx, mondooAuditConfigs); err != nil {
+		h.Log.Error(err, "error listing MondooAuditConfigs")
+		return err
+	}
+
+	for _, mac := range mondooAuditConfigs.Items {
+		if mac.Status.ReconciledByOperatorVersion != version.Version {
+			errorMsg := fmt.Sprintf("MondooAuditConfig not yet completly reconciled: namespace=%s MondooAuditConfig=%s", mac.Namespace, mac.Name)
```
imilchev, line 30: To make the message more descriptive you can rephrase it to something like: "MondooAuditConfig 'ns'/'name' has not been reconciled by operator 'version'." You could even state what is the current value of the `ReconciledByOperatorVersion` field

pkg/health/health.go
```diff
@@ -0,0 +1,40 @@
...
+			h.Log.Info(errorMsg)
+			return errors.New(errorMsg)
+		} else {
+			infoMsg := fmt.Sprintf("MondooAuditconfig reconciled: namespace=%s MondooAuditConfig=%s", mac.Namespace, mac.Name)
+			h.Log.Info(infoMsg)
```
imilchev, line 35: I think this should be a debug message since it isn't really required by the end-user to know this

tests/integration/audit_config_base_suite.go
```diff
@@ -408,3 +448,18 @@ func (s *AuditConfigBaseSuite) checkForPodInStatus(podName string) error {
 
 	return err
 }
+
+// checkForReconciledOperatorVersion Check whether the MondooAuditConfig Status contains the current operator Version after Reconcile.
+func (s *AuditConfigBaseSuite) checkForReconciledOperatorVersion() error {
```
imilchev, line 453: you could move this function somewhere in the test framework instead of defining it in the test suite. In this way it can be re-used by other suites in the future. If you choose not to do it, then you can make it `void` as it doesn't need to return the error. You can already assert in this function

tests/integration/audit_config_base_suite.go
```diff
@@ -298,6 +308,36 @@ func (s *AuditConfigBaseSuite) testMondooAuditConfigAdmissionMissingSA(auditConf
...
+		s.testCluster.K8sHelper.Clientset.Create(s.ctx, operatorConfig), "Failed to create MondooOperatorConfig")
+
+	// Enable nothing
+	zap.S().Info("Create an audit config that enables nothing.")
+	s.NoErrorf(
+		s.testCluster.K8sHelper.Clientset.Create(s.ctx, &s.auditConfig),
+		"Failed to create Mondoo audit config.")
+
+	err := s.checkForReconciledOperatorVersion()
+	s.Assert().NoErrorf(err, "Couldn't find expected version in MondooAuditConfig.Status.ReconciledByOperatorVersion")
```
imilchev, line 340: I believe `s.Assert().NoErrorf(..)` is equivalent to `s.NoErrorf(..)` or is that not the case?

czunker: You are correct. Changed all occurrences.

tests/integration/audit_config_base_suite.go
czunker: I moved all the functions similar to this one into the framework.

pkg/health/health.go
czunker: As there is no debug level, I removed it completely.

pkg/health/health.go
czunker: Changed.

pkg/health/health.go
czunker: Changed

pkg/health/health.go
czunker: Done.

imilchev (approved):
looks great! let's ship it

controllers/mondooauditconfig_controller.go
```diff
@@ -108,6 +110,7 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	mondooAuditConfig := &v1alpha2.MondooAuditConfig{}
 
 	err := r.Get(ctx, req.NamespacedName, mondooAuditConfig)
+	r.MondooAuditConfig = mondooAuditConfig
```
joelddiaz, line 113: It appears to me that this is only used for test case assertions. This should be fine for a single-threaded execution model (which is how our tests currently run), but we should not rely on this for any runtime-related activities as I would expect the Reconcile() function to have to be written in a re-entrant way.

PR #396, opened 2022-06-21
🧹 Refactor parts of integration tests

czunker:
- Revert kustomization.yaml to previous state after applying them
  The files contained the image tag with the current git hash
- Suppress stack traces during log collection
  Sometimes there are no logs to collect

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,15 +150,21 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	cp config/manager/kustomization.yaml config/manager/kustomization.yaml.before_kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	mv config/manager/kustomization.yaml.before_kustomize config/manager/kustomization.yaml
 
-.PHONY: undeploy
+.PHONY: generate-manifests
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -150,15 +150,21 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	cp config/manager/kustomization.yaml config/manager/kustomization.yaml.before_kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	mv config/manager/kustomization.yaml.before_kustomize config/manager/kustomization.yaml
```
imilchev, line 156: that's very neat! Finally, we don't need to remember to not commit these files

Makefile
```diff
@@ -197,7 +203,7 @@ GOMOCKGEN = $(shell pwd)/bin/mockgen
 .PHONY: gomockgen
 gomockgen: ## Download go mockgen locally if necessary.
 	$(call go-get-tool,$(GOMOCKGEN),github.com/golang/mock/mockgen@v1.6.0)
-	# mockgen binary needs to be in $PATH
+	# mockgen binary needs to be in $$PATH
```
imilchev, line 206: why the double `$$`

imilchev (approved):

Makefile
czunker: The comment was printed as
```
# mockgen binary needs to be in ATH
```
After seeing it once, I saw it every time and had to change it.

It's just formatting.

PR #399, opened 2022-06-22
Add mockgen to $PATH without copy

czunker:
`/usr/local/bin` isn't writeable by default by an unprivileged user on Ubuntu.
So, instead of copying the file, add the directory to $PATH.


czunker:
@joelddiaz / @imilchev Please double check whether this also works in your environments.

imilchev:
Try first if #401 does the trick for Linux as well. For Mac it works just fine

czunker:
Is fixed by #401

PR #404, opened 2022-06-23
🧹 Fix typos and staticchecks

czunker:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,7 +204,7 @@ golangci-lint: ## Download golangci-lint locally if necessary.
 
 GO_STATICCHECK = $(shell pwd)/bin/staticcheck
 go-staticcheck: ## Download go-staticcheck locally if necessary.
-	$(call go-get-tool,$(GO_STATICCHECK),honnef.co/go/tools/cmd/staticcheck@latest)
+	$(call go-install-tool,$(GO_STATICCHECK),honnef.co/go/tools/cmd/staticcheck@latest)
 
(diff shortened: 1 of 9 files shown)


imilchev:
can we also make sure the linting step catches these cases? In this way we will make sure we do not let this happen in the first place in our CI

Makefile
```diff
@@ -199,6 +202,10 @@ GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 golangci-lint: ## Download golangci-lint locally if necessary.
 	$(call go-get-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/cmd/golangci-lint@v1.45)
 
+GO_STATICCHECK = $(shell pwd)/bin/staticcheck
+go-staticcheck: ## Download go-staticcheck locally if necessary.
```
imilchev, line 206: golagci-lint already supports staticcheck. See [here](https://golangci-lint.run/usage/linters/#staticcheck). Can we use that?

czunker: Not now, that we updated to go1.18:
```
WARN [linters context] staticcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
```
I subscribed to the issue. Perhaps we can update in the future.

czunker: https://github.com/mondoohq/mondoo-operator/issues/405

imilchev (approved):
LGTM

PR #407, opened 2022-06-24
🐛 Remove version check from Readniess check

### LINKED ISSUE #406: Deadlock when upgrading the operator to v0.4.1
The v0.4.1 of the operator adds the feature for checking whether all `MondooAuditConfig` have been reconciled by the latest version. However, this result in a deadlock if the user already has an older version of the operator running in the cluster. We use leader election to make sure there is only 1 operator instance actively reconciling CRDs. If we upgrade a running operator the old version of the operator will be active until the pod with the new version becomes ready. In order for the new pod to be come ready it needs the lock to be released by the old version so it can actually reconcile the CRDs.

czunker:
This brakes operator updates.
Now, the docs include a command to manually check during operator updates.

Fixes #406


docs/operator-upgrades.md
```diff
@@ -33,6 +33,13 @@ If there **is** more than one minor version difference between the installed Mon
     kubectl logs -n mondoo-operator deployment/mondoo-operator-controller-manager
     ```
     It's essential to check the logs and wait for the new version of the operator to run; directly upgrading to the next version can result in skipped internal upgrade procedures and unexpected behavior.
+
+    Check the `Status.ReconciledByOperatorVersion` field (introduced with v0.4.1)  of your `MondooAuditConfig` to be sure the new operator version reconciled every object:
+    ```bash
+    kubectl get mondooauditconfigs.k8s.mondoo.com -n mondoo-operator mondoo-client -o jsonpath='{.status.reconciledByOperatorVersion}'
```
imilchev, line 39: let's change this command to list all CRDs -> select the `reconciledByOperatorVersion` -> print only the distinct version

If we do that, then we can say that this command should return only the version that was just installed and nothing older

cmd/mondoo-operator/operator/cmd.go
```diff
@@ -101,11 +100,7 @@ func init() {
 			return err
 		}
 
-		reconcileCheck := &health.HealthChecks{
```
imilchev, line 104: do we also want to delete the `AreAllMondooAuditConfigsReconciled`? If we keep it, then we need to maintain that piece of code as well as we migrate CRD versions

czunker: Good point, missed the maintenance for the CRD migration.

docs/operator-upgrades.md
czunker: Changed.

docs/operator-upgrades.md
```diff
@@ -33,6 +33,13 @@ If there **is** more than one minor version difference between the installed Mon
     kubectl logs -n mondoo-operator deployment/mondoo-operator-controller-manager
     ```
     It's essential to check the logs and wait for the new version of the operator to run; directly upgrading to the next version can result in skipped internal upgrade procedures and unexpected behavior.
+
+    Check the `Status.ReconciledByOperatorVersion` field (introduced with v0.4.1)  of your `MondooAuditConfig` to be sure the new operator version reconciled every object:
+    ```bash
+    kubectl get mondooauditconfigs.k8s.mondoo.com -o jsonpath='{.items[*].status.reconciledByOperatorVersion}'
```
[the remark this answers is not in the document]
czunker: Done

imilchev (approved):

PR #408, opened 2022-06-27
⭐ Set Admission Webhook `failurePolicy` depending ...

### LINKED ISSUE #328: Investigate admission controller error behaviour
A few times I encountered some errors in the admission controller that we deploy. I still don't know why but sometimes it was getting weird response from the scan API and then it fails. I only saw that by accident because in case an error like this happens, right now we just allow the resource. I understand that we probably implemented it like this so we are not intrusive in case our stuff doesn't work. However, I do see why this might turn out to be not an ideal solution. I think it might be better to just deny everything until the errors in the admission controller are resolved. If I have enabled the admission controller, I would expect it to do its work. In case it cannot because of misconfiguration or because of a bug my deployments should be blocked so at least I am aware of the issue. At that point I can decide whether I want to fix the problem or disable the admission controller.

czunker:
... on the mode of the webhook.
This also implies changing the responses depending on the mode.

In case the replica isn't expicitly set, we should also scale to two
replicas with `PodAntiAffinity` to prevent some failure modes.

Fixes #328


czunker:
Still work in progress. What's missing:

- [ ] docs
- [ ] unit tests
- [ ] integration tests

czunker:
With the current implementation, it is not possible to set admission mode to `enforcing` and the ValidationWebhook `failurePolicy` to ignore. Not sure, whether anyone would run this combination. To me, it doesn't make much sense.

api/v1alpha2/mondooauditconfig_types.go
```diff
@@ -60,6 +60,11 @@ type Scanner struct {
 	ServiceAccountName string                      `json:"serviceAccountName,omitempty"`
 	Image              Image                       `json:"image,omitempty"`
 	Resources          corev1.ResourceRequirements `json:"resources,omitempty"`
+	// Number of replicas for the scanner. When not set and admission mode is "enforcing", it will
```
imilchev, line 63: why set it to 2? That means for tiny clusters we are going to be using a lot of resources. I suspect you did this because otherwise it might happen that the admission controller blocks its own bootstrapping. With the changes I did in #414 that will no longer be the case. I think we should have it as 1 for enforcing as well. I do like the fact that we allow for changing the replicas though so we should definitely keep that new propery.

controllers/admission/deployment_handler.go
```diff
@@ -221,6 +235,10 @@ func (n *DeploymentHandler) syncWebhookDeployment(ctx context.Context) error {
 		return err
 	}
 
+	if *desiredDeployment.Spec.Replicas < 2 && n.Mondoo.Spec.Admission.Mode == mondoov1alpha2.Enforcing {
```
imilchev, line 238: I think you accept my other comment this specific check wouldn't be needed but we should definitely validate that the replicas cannot be less than 1

controllers/admission/deployment_handler.go
```diff
@@ -221,6 +235,10 @@ func (n *DeploymentHandler) syncWebhookDeployment(ctx context.Context) error {
...
+		webhookLog.Info("WARNING: Webhook deployment is only scaled to 1 replica, but the webhook mode is set to 'enforcing'. This might be problematic if the API server is not able to connect to the webhook. Please consider increasing the replicas.")
```
imilchev, line 239: doesn't this logger have a `Warn` func?

controllers/admission/deployment_handler.go
```diff
@@ -247,6 +271,23 @@ func (n *DeploymentHandler) syncWebhookDeployment(ctx context.Context) error {
...
+	for _, condition := range n.Mondoo.Status.Conditions {
+		if condition.Type == mondoov1alpha2.ScanAPIDegraded {
+			if condition.Status == corev1.ConditionTrue {
+				webhookLog.Info("Webhook is degraded, becasue ScanAPI is degraded. Please check the ScanAPI status.")
+				return true
+			}
+		}
+	}
+	if reflect.DeepEqual(deployment.Status, appsv1.DeploymentStatus{}) {
+		webhookLog.Info("No status present for webhook deployment. Assuming it is starting.")
```
imilchev, line 284: didn't you create a PR recently to switch all log statements to start with a small letter to follow the convention?

api/v1alpha2/mondooauditconfig_types.go
```diff
@@ -78,7 +83,12 @@ type Admission struct {
 	// on the scan results may reject the k8s resource creation/modification.
 	// +kubebuilder:validation:Enum=permissive;enforcing
 	// +kubebuilder:default=permissive
-	Mode                    AdmissionMode           `json:"mode,omitempty"`
+	Mode AdmissionMode `json:"mode,omitempty"`
+	// Number of replicas for the admission webhook. When not set and mode is "enforcing", it will
+	// automtically be set to two. When not set and mode is "permissive, it will be set to one.
+	// For enforcing mode, the minimum should be two to prevent problems during Pod failures,
+	// e.g. node failure, node scaling, etc.
+	Replicas                *int32                  `json:"replicas,omitempty"`
```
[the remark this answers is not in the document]
imilchev: actually for the admission controller itself it might be a problem if it crashes, so this one can be 2 replicas (it is also very minimal on resource requirements). With my changes if the admission controller is running it will make sure it skips pods with a parent but if there is no admission controller, then we will be blocking ourselves

controllers/scanapi/conditions.go
```diff
@@ -31,4 +31,16 @@ func updateScanAPIConditions(config *mondoov1alpha2.MondooAuditConfig, degradedS
 	}
 
 	config.Status.Conditions = mondoo.SetMondooAuditCondition(config.Status.Conditions, mondoov1alpha2.ScanAPIDegraded, status, reason, msg, updateCheck)
+	if degradedStatus {
+		// also admission is degraded
+		for _, condition := range config.Status.Conditions {
```
imilchev, line 36: I think this is an inversed ownership. From the perspective of the scan API it is not important that there is an admission controller. Therefore it is weird that the scan API logic is updating the condition for the admission controller. If we want this, I think the logic for doing that should be in the admission controller

controllers/scanapi/deployment_handler.go
```diff
@@ -111,6 +111,11 @@ func (n *DeploymentHandler) syncDeployment(ctx context.Context) error {
 	if err := ctrl.SetControllerReference(n.Mondoo, deployment, n.KubeClient.Scheme()); err != nil {
 		return err
 	}
+
+	if *deployment.Spec.Replicas < 2 && n.Mondoo.Spec.Admission.Mode == v1alpha2.Enforcing {
+		logger.Info("WARNING: Scan API deployment is only scaled to 1 replica, but the admission mode is set to 'enforcing'. This might be problematic if the API server is not able to connect to the admission webhook. Please consider increasing the replicas.")
```
imilchev, line 116: just read about `logr.Logger`. Apparently we are supposed to use a verbosity level for these cases. The higher the level of verbosity, the more logging we are spitting out. This means the important stuff should be at a lower verbosity level than the "debug" info.

controllers/admission/deployment_handler.go
czunker: The convention is strange and only applies to error messages.

controllers/scanapi/deployment_handler.go
czunker: Yes, but could you find out the V(?) equivalent to the debug or info level? I couldn't find anything useful.

Perhaps a warning might be `V(3)`?

api/v1alpha2/mondooauditconfig_types.go
czunker: The two replicas aren't for the bootstrap, but for the case, that when the webhook is already enforcing the policies and then an error occurs, like a node failure or something like that.

Until the Pod is rescheduled (hopefully it happens) the webhook would block requests.

controllers/scanapi/deployment_handler.go
imilchev: It's up to our own interpretation. We have to pick our own numbering

api/v1alpha2/mondooauditconfig_types.go
```diff
@@ -60,6 +60,11 @@ type Scanner struct {
 	ServiceAccountName string                      `json:"serviceAccountName,omitempty"`
 	Image              Image                       `json:"image,omitempty"`
 	Resources          corev1.ResourceRequirements `json:"resources,omitempty"`
+	// Number of replicas for the scanner. When not set and admission mode is "enforcing", it will
+	// automtically be set to two. When not set and admission mode is "permissive, it will be set to one.
+	// For enforcing mode, the minimum should be two to prevent problems during Pod failures,
+	// e.g. node failure, node scaling, etc.
+	Replicas *int32 `json:"replicas,omitempty"`
```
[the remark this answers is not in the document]
czunker: I intentionally left the default out.
This way, I can decide in the code, whether the user wants to overwrite the 2 replicas default for enforcing mode.

controllers/scanapi/conditions.go
czunker: Yes, tried it the other way around, but this didn't work in the tests.
Idea was:
- Scan API is down
- controller would update the status of the Scan API in the MondooAuditConfig
- this would trigger the reconcile for the other controllers
- Admission controller would update its own Status to degraded, because the Scan API is degraded

But it seems, the update to the Status of the MondooAuditConfig didn't trigger the reconcile.

In case I missed something, I'd like to move the above code somewhere else.

controllers/admission/deployment_handler.go
```diff
@@ -247,6 +271,23 @@ func (n *DeploymentHandler) syncWebhookDeployment(ctx context.Context) error {
 	return nil
 }
 
+func (n *DeploymentHandler) isWebhookDegraded(deployment *appsv1.Deployment) bool {
+	for _, condition := range n.Mondoo.Status.Conditions {
+		if condition.Type == mondoov1alpha2.ScanAPIDegraded {
+			if condition.Status == corev1.ConditionTrue {
```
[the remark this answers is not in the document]
czunker: Good catch.

api/v1alpha2/mondooauditconfig_types.go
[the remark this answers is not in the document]
czunker: Added annotation

controllers/scanapi/deployment_handler.go
czunker: Changed.

controllers/admission/deployment_handler.go
```diff
@@ -247,6 +271,21 @@ func (n *DeploymentHandler) syncWebhookDeployment(ctx context.Context) error {
 	return nil
 }
 
+func (n *DeploymentHandler) isWebhookDegraded(deployment *appsv1.Deployment) bool {
+	for _, condition := range n.Mondoo.Status.Conditions {
```
imilchev, line 275: there is a function `mondoo.FindMondooAuditConditions`

api/v1alpha2/mondooauditconfig_types.go
czunker: Changed code and docs accordingly.

controllers/admission/deployment_handler.go
czunker: Used it. Thanks for the hint.

tests/integration/audit_config_base_suite.go
```diff
@@ -390,3 +434,167 @@ func (s *AuditConfigBaseSuite) disableContainerImageResolution() func() {
...
+						ProbeHandler: corev1.ProbeHandler{
+							Exec: &corev1.ExecAction{
+								Command: []string{"/bin/sh", "-c", "exit 0"},
+							},
+						},
+					},
+				},
+			},
+			AutomountServiceAccountToken: pointer.Bool(false),
+			ServiceAccountName:           "mondoo-operator-controller-manager",
```
joelddiaz, line 490: Why set the ServiceAccountName if AutomountServiceAccountToken is false?

controllers/admission/deployment_handler_test.go
```diff
@@ -74,6 +75,114 @@ func TestReconcile(t *testing.T) {
...
+				mac.Admission.Replicas = pointer.Int32(2)
+				return mac
+			}(),
+			validate: func(t *testing.T, kubeClient client.Client) {
+				deployment := &appsv1.Deployment{}
+				deploymentKey := types.NamespacedName{Name: webhookDeploymentName(testMondooAuditConfigName), Namespace: testNamespace}
+				err := kubeClient.Get(context.TODO(), deploymentKey, deployment)
+				require.NoError(t, err, "expected Admission Deployment to exist")
+
+				// when replicas is set, the deployment should have a replicas count of 1, although it is enforcing
```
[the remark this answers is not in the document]
czunker: The comment was useless. Thanks.

controllers/mondooauditconfig_controller.go
```diff
@@ -175,6 +176,41 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		// Update status.ReconciledByOperatorVersion to the running operator version
+		if r.MondooAuditConfig.Status.ReconciledByOperatorVersion != version.Version {
+			r.MondooAuditConfig.Status.ReconciledByOperatorVersion = version.Version
+		}
+
+		funcErr := mondoo.UpdateMondooAuditStatus(ctx, r.Client, mondooAuditConfigCopy, r.MondooAuditConfig, log)
+		// do not overwrite errors which happened before the defered function is called
+		if err == nil {
+			err = funcErr
+		}
```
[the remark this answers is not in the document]
czunker: This is all part of a defer block. So, err might already contain errors from webhook reconcile or something like this.
I'd prefer to see these, before a "pod list" error masks these.

controllers/scanapi/deployment_handler.go
```diff
@@ -106,11 +105,18 @@ func (n *DeploymentHandler) syncDeployment(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	logger.Info("Mondoo client image: ", "image", mondooClientImage)
+	logger.Info("Mondoo skip resolve: ", "SkipContainerResolution", n.MondooOperatorConfig.Spec.SkipContainerResolution)
```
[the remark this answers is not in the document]
czunker: Changed.

tests/integration/audit_config_base_suite.go
czunker: Shouldn't be needed anymore. Policy bugifx is released.

controllers/admission/deployment_handler.go
```diff
@@ -281,6 +314,14 @@ func (n *DeploymentHandler) prepareValidatingWebhook(ctx context.Context, vwc *w
 		annotationValue = "manual"
 	}
 
+	for i := range vwc.Webhooks {
+		if n.Mondoo.Spec.Admission.Mode == mondoov1alpha2.Enforcing {
+			*vwc.Webhooks[i].FailurePolicy = webhooksv1.Fail
+		} else {
+			*vwc.Webhooks[i].FailurePolicy = webhooksv1.Ignore
+		}
+	}
```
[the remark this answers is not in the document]
czunker: I'm not sure about this.

I get your point of making t hard for the user to not shoot himself in the foot. But, by splitting this up we would have two places where we configure a part of the deployed objects.

One other thing is resource usage. Let's assume we really have multiple MondooAuditConfigs, and only one has the admission mode set to enforcing, all would still get the replicas specified in a central MondooOperatorConfig.

Ok, writing this, another scenario might be multiple Admission webhooks deployed to the same cluster. This would chain them together, making it take a long time for any request to complete. This is another way users might shoot themself in the foot.

There should only be one Admission webhook per cluster. Do we really need to have the MondooAuditConfig namespaced? Or should we only search for MondooAuditConfigs in the namespace where the operator is deployed?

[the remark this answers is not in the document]
imilchev: Replicas = 0 shouldn't be possible because we restrict that. I think we are forgetting the fact that different audit configs can actually have different Mondoo spaces and service accounts which means different policies. I think it is applicable that 1 space might be just observing and another might have to be enforcing. I think the global Permissive/Enforcing is a little restricting

controllers/mondooauditconfig_controller.go
```diff
@@ -175,6 +176,41 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		ctx := context.Background()
+		// Update the mondoo status with the pod names only after all pod creation actions are done
+		// List the pods for this mondoo's cronjobs and deployment
+		podList := &corev1.PodList{}
+		listOpts := []client.ListOption{
+			client.InNamespace(mondooAuditConfig.Namespace),
+			client.MatchingLabels(labelsForMondoo(mondooAuditConfig.Name)),
+		}
+		if funcErr := r.List(ctx, podList, listOpts...); funcErr != nil {
+			log.Error(funcErr, "Failed to list pods", "Mondoo.Namespace", mondooAuditConfig.Namespace, "Mondoo.Name", mondooAuditConfig.Name)
```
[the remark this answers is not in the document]
joelddiaz: Ah, I think I missed that the funcErr here is scoped to the `if` block...

Looking at it again, if I'm reading the flow correctly I see this possibility:

`r.List()` for the Pods errors, and presumably `podList` is an empty list; meaning `podListNames` is returned as an empty set.

It then compares the existing list of Pods in the MondooAuditConfig.Status (as a set comparison) and sees that the sets aren't equal (assuming there was a list of Pods previously in the Status). This would then wipe out the previous list of Pods in the Status.

I assume this isn't intended behavior??? So does it make sense to put the Pod set/list comparison as an `else` clause for the `funcErr == nil` case? This way the Pod list in the Status is only updated if there was no error listing the Pods.

controllers/admission/deployment_handler.go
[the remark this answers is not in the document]
joelddiaz: After chatting about this in person, I see that the Replicas wouldn't be zero in any case.

And I think we agreed that we don't need to solve the "rogue/misbehaving MondooAuditConfig" in this PR, but we should keep in mind some way to mitigate potential breakage due to a Webhook Fail policy as we get closer to being able to consider enforcing mode a fully supported mode of operation.

But for the purposes of this PR, this should be okay.

controllers/mondooauditconfig_controller.go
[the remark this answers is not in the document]
joelddiaz: I see the goal here. In the past when doing this kind of pattern I've found it helpful to set pre-named return variables.
I'm not saying it's a requirement to use named variables here, but I would suggest considering it so that rather than the code:
```
if err == nil {
  err = funcErr
}
```

We could end up with something that reads like:

```
if reconcileErr == nil {
  reconileErr = funcErr
}
```

controllers/mondooauditconfig_controller.go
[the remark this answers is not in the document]
czunker: Good catch, thanks.

joelddiaz (approved):

imilchev (approved):
LGTM

PR #409, opened 2022-06-27
Release 0.4.3

czunker:

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=mondoo-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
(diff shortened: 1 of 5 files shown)


bundle.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=mondoo-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.19.0+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.21.0
```
imilchev, line 9: there is this change again. Can you verify what needs to be done to upgrade to the latest version of the operator SDK. Just check what we currently use and follow the steps (if any). Before the blocker for migrating was Go 1.18. Now you fixed it so we can also update this. The guide is [here](https://sdk.operatorframework.io/docs/upgrading-sdk-version/)

imilchev (approved):

PR #414, opened 2022-06-30
⭐ Filter out resources which owners we have already scanned for admission

### LINKED ISSUE #413: Skip scanning of pods created from resources we scan
We scan a LOT. So much that we bury the useful scans.

Avoid double scanning and instead only scan the resources that later create the pods.

UI filtering, but that doesn't avoid all the unnecessary compute and potential for blocking pod creation that we have if we keep scanning the pods.

imilchev:
Closes #413 

- [x] Filter out resources
- [x] Add unit tests
- [x] Make sure we scan `CronJob`, `Job`, `StatefulSet` with the admission controller


tas50:
+1 from the product side here

pkg/webhooks/handler/webhook.go
```diff
@@ -1,21 +1,24 @@
...
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	"sigs.k8s.io/yaml"
 
+	serializerYaml "k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+
 	mondoov1alpha2 "go.mondoo.com/mondoo-operator/api/v1alpha2"
 	"go.mondoo.com/mondoo-operator/pkg/constants"
 	"go.mondoo.com/mondoo-operator/pkg/mondooclient"
-	customobject "go.mondoo.com/mondoo-operator/pkg/utils/genericobjectdecoder"
```
czunker, line 18: Haven't found another reference for this.
Guess we can then remove it.

czunker (approved):

pkg/webhooks/handler/webhook.go
imilchev: done

PR #415, opened 2022-06-30
🐛 Revert to `go install` for kustomize

czunker:
The new `curl | bash` command for operator-sdk 1.21.0 does not work probably, when executed mutliple times.
`go install` takes care of this and is alligned with the other tools we install.


imilchev (approved):

joelddiaz:
This looks like fallout from diverging from the version of kustomize used by OperatorSDK. Perhaps commenting out the original OperatorSDK Makefile lines with a comment on why we are different for our future selves to remember what is happening here...

Issue #394, opened 2022-06-20
Report additional information to the Mondoo Backend


czunker:
**Describe the solution you'd like**
Report the current operator version and MondooAuditConfig to the Mondoo Backend.

**Describe alternatives you've considered**
-

**Additional context**
Currently, the operator reports back to Mondoo Backend every 10 minutes. But it does not report the above-mentioned details. So, the data gets out of sync over time.

PR #419, opened 2022-07-01
⭐ Report integration status

imilchev:
- [x] Report integration status
- [x] Add unit tests

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "gopls": {
+        "formatting.gofumpt": true,
+    },
+}
--- a/cmd/mondoo-operator/operator/cmd.go
+++ b/cmd/mondoo-operator/operator/cmd.go
@@ -23,6 +23,7 @@ import (
 	"go.mondoo.com/mondoo-operator/controllers"
 	"go.mondoo.com/mondoo-operator/controllers/integration"
 	"go.mondoo.com/mondoo-operator/controllers/status"
+	"go.mondoo.com/mondoo-operator/pkg/utils/k8s"
 	"go.mondoo.com/mondoo-operator/pkg/utils/mondoo"
 	"go.mondoo.com/mondoo-operator/pkg/version"
 	//+kubebuilder:scaffold:imports
@@ -74,11 +75,17 @@ func init() {
 			return err
 		}
 
+		v, err := k8s.GetServerVersion(mgr.GetConfig())
+		if err != nil {
+			setupLog.Error(err, "failed to retrieve server version")
+			return err
+		}
+
 		if err = (&controllers.MondooAuditConfigReconciler{
 			Client:                 mgr.GetClient(),
 			MondooClientBuilder:    controllers.MondooClientBuilder,
 			ContainerImageResolver: mondoo.NewContainerImageResolver(),
-			StatusReporter:         status.NewStatusReporter(mgr.GetClient(), controllers.MondooClientBuilder),
+			StatusReporter:         status.NewStatusReporter(mgr.GetClient(), controllers.MondooClientBuilder, v),
 		}).SetupWithManager(mgr); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "MondooAuditConfig")
 			return err
@@ -106,7 +113,7 @@ func init() {
 			return err
 		}
 
-		setupLog.Info("starting manager", "version", version.Version)
+		setupLog.Info("starting manager", "operator-version", version.Version, "k8s-version", v.GitVersion)
 		if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
 			setupLog.Error(err, "problem running manager")
 			return err
--- a/cmd/mondoo-operator/operator/cmd.go
+++ b/cmd/mondoo-operator/operator/cmd.go
@@ -22,6 +22,7 @@ import (
 	k8sv1alpha2 "go.mondoo.com/mondoo-operator/api/v1alpha2"
 	"go.mondoo.com/mondoo-operator/controllers"
 	"go.mondoo.com/mondoo-operator/controllers/integration"
+	"go.mondoo.com/mondoo-operator/controllers/status"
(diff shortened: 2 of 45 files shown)


czunker:
This fixes #394

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "gopls": {
```
imilchev, line 2: we already use this setting in the Mondoo repo. It does change the formatting a bit. I am adding it here so we share the settings

tests/framework/utils/k8s_helper.go
```diff
@@ -334,7 +333,7 @@ func (k8sh *K8sHelper) createTestLogFile(name, namespace, testName, suffix strin
 	dir, _ := os.Getwd()
 	logDir := path.Join(dir, "_output/tests/")
 	if _, err := os.Stat(logDir); os.IsNotExist(err) {
-		err := os.MkdirAll(logDir, 0777)
+		err := os.MkdirAll(logDir, 0o777)
```
czunker, line 336: Why this change?

I also saw it in some other files for the `DefaultMode` of secrets.

czunker:
Nice! :tada:

czunker (approved):

pkg/mondooclient/client.go
```diff
@@ -364,3 +381,65 @@ func (s *mondooClient) IntegrationCheckIn(ctx context.Context, in *IntegrationCh
 
 	return out, nil
 }
+
+type ReportStatusRequest struct {
+	Mrn string `protobuf:"bytes,1,opt,name=mrn,proto3" json:"mrn,omitempty"`
+	// this is the status of the integration itself (is it active/checking in, errored, etc)
+	Status Status `protobuf:"varint,2,opt,name=status,proto3,enum=mondoo.integrations.v1.Status" json:"status,omitempty"`
+	// this can be any information about the current state of the integration. it will be displayed to the user as-is where supported
+	LastState interface{} `protobuf:"bytes,4,opt,name=last_state,json=lastState,proto3" json:"last_state,omitempty"`
```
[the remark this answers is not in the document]
imilchev: this is copy-pasted code from the mondoo repo because we cannot generate those files ourselves at the moment. Ignore all the types defined here

pkg/mondooclient/client.go
```diff
@@ -364,3 +381,65 @@ func (s *mondooClient) IntegrationCheckIn(ctx context.Context, in *IntegrationCh
...
+	// Allows the agent to report its current version
+	Version string `protobuf:"bytes,5,opt,name=version,proto3" json:"version,omitempty"`
+	// messages that convey extra information about the integration - these messages can be informational, warnings or errors. Can be used
+	// to report non-critical errors/warnings without neccesarily changing the whole integration status.
+	Messages []IntegrationMessage `protobuf:"bytes,7,opt,name=messages,proto3" json:"messages,omitempty"`
```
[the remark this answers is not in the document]
imilchev: same

tests/framework/utils/k8s_helper.go
imilchev: I believe the `gofumpt` linter does that because the other notation is not accepted

imilchev:
@joelddiaz since you did the console integration initially it might be good to have a look at this PR as well to make sure I haven't missed something

token
```diff
@@ -0,0 +1 @@
+abcdefgh
```
joelddiaz, line 1: This some leftover local debug/test file?

joelddiaz (commented):
One thing we've discussed previously was sending back either the full or a piece of the MondooAuditConfig so that the UI can do things like say "your last reported configuration differs from the configuration stored in Mondoo".

Assuming that is still desired, we may want to send along enough of the MondooAuditConfig so that we can do a server-side determination when the running config doesn't match the desired config.

Also, (maybe a question for @vjeffrey ): I was under the impression that we could use the ReportStatus() call instead of the CheckIn() call. Therefore I just expected the CheckIn()s to basically go away and be replaced by ReportStatus() calls instead. Not that the approach in the PR is wrong, but just for my knowledge, is there any particular reason do keep the CheckIn() as-is and adding this new codepath with the ReportStatus() calls?

imilchev:
> One thing we've discussed previously was sending back either the full or a piece of the MondooAuditConfig so that the UI can do things like say "your last reported configuration differs from the configuration stored in Mondoo".
> 
> 
> 
> Assuming that is still desired, we may want to send along enough of the MondooAuditConfig so that we can do a server-side determination when the running config doesn't match the desired config.
> 
> 
> 
> Also, (maybe a question for @vjeffrey ): I was under the impression that we could use the ReportStatus() call instead of the CheckIn() call. Therefore I just expected the CheckIn()s to basically go away and be replaced by ReportStatus() calls instead. Not that the approach in the PR is wrong, but just for my knowledge, is there any particular reason do keep the CheckIn() as-is and adding this new codepath with the ReportStatus() calls?

I can explain the reasoning behind this. You are right that ReportStatus implicitly does a CheckIn as well. However, here we still use both because I didn't want to just replace the CheckIn with ReportStatus and report every 10mins. The current implementation  checks in every 10mins and does ReportStatus every time the status actually changes. If there is no change in the status (e.g. config hasn't changed and no errors have occurred), we do not report the status object. For that reason the checkin logic is still there, such that we still do a ping even if there are no status changes

cmd/mondoo-operator/operator/cmd.go
```diff
@@ -74,11 +75,17 @@ func init() {
 			return err
 		}
 
+		v, err := k8s.GetServerVersion(mgr.GetConfig())
```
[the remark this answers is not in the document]
imilchev: that is also my understanding so I implemented it like this. Basically is you do a full cluster upgrade the operator will be re-scheduled at least once for sure, which means we will get the new version. If you upgrade half of your nodes, then that won't happen but then you are also doing something weird, so I wouldn't try to fix that situation

imilchev:
Processed the comments and rebased on `main`

joelddiaz (approved):

PR #421, opened 2022-07-04
⭐ Add Resource to Nodes type

### LINKED ISSUE #390: Support setting resources requests/limits for node scanning
Currently, setting resource requests/limits is set on `MondooAuditConfig.Scanner.Resources`. That property is used by both the Scan API and node scanning. Originally the plan was to re-use the Scan API for doing node scans but we are long way to go from there. We should allow setting resource requests/limits for node scanning separately to have more granular control over what is deployed.

czunker:
Nodes resource requirements can now be configured seperatly.

Fixes #390


imilchev (approved):
LGTM

PR #430, opened 2022-07-05
⭐ Container image scanning

imilchev:
Allow scanning container images that are used by the workloads in the cluster. Currently, this is implemented as a new `CronJob` that runs once per day because for large clusters container scanning can take a while.

- [x] implement change
- [x] extend tests
- [x] ~~document use-case~~ -> will be done in the docs repo and not here

### DIFF
--- a/controllers/k8s_scan/container_image/deployment_handler.go
+++ b/controllers/k8s_scan/container_image/deployment_handler.go
@@ -14,32 +14,22 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package k8s_scan
+package container_image
 
 import (
 	"context"
-	"fmt"
 
-	appsv1 "k8s.io/api/apps/v1"
+	"go.mondoo.com/mondoo-operator/api/v1alpha2"
+	"go.mondoo.com/mondoo-operator/pkg/utils/k8s"
+	"go.mondoo.com/mondoo-operator/pkg/utils/mondoo"
 	batchv1 "k8s.io/api/batch/v1"
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"go.mondoo.com/mondoo-operator/api/v1alpha2"
-	"go.mondoo.com/mondoo-operator/controllers/k8s_scan/container_image"
-	"go.mondoo.com/mondoo-operator/pkg/utils/k8s"
-	"go.mondoo.com/mondoo-operator/pkg/utils/mondoo"
 )
 
-var logger = ctrl.Log.WithName("k8s-resources-scanning")
-
-const (
-	workloadDeploymentConfigMapNameTemplate = `%s-deploy`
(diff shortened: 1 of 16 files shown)


controllers/k8s_scan/container_image/resources.go
```diff
@@ -0,0 +1,133 @@
...
+											corev1.ResourceMemory: resource.MustParse("100Mi"),
+										},
+										Requests: corev1.ResourceList{
+											corev1.ResourceCPU:    resource.MustParse("100m"),
+											corev1.ResourceMemory: resource.MustParse("20Mi"),
+										},
+									},
+									SecurityContext: &corev1.SecurityContext{
+										AllowPrivilegeEscalation: pointer.Bool(false),
+										ReadOnlyRootFilesystem:   pointer.Bool(true),
```
czunker, line 91: Where are the images stored for the scan? Is it in memory or somewhere on disk?

When it is in-memory: Can we scan bigger images with these memory limits?

As an example for a bigger image: https://hub.docker.com/r/jupyter/datascience-notebook/tags

imilchev: the tar for an image is stored in `/tmp` and is removed after the scan. It shouldn't be a problem even for this large image as long as there is sufficient disk space

czunker: Ah, I see via the Scan API. Thanks. :+1:

controllers/k8s_scan/container_image/resources.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if integrationMrn != "" {
+		containerArgs = append(containerArgs, []string{"--integration-mrn", integrationMrn}...)
+	}
+
+	return &batchv1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      CronJobName(m.Name),
+			Namespace: m.Namespace,
+			Labels:    CronJobLabels(m),
```
czunker, line 57: You could use `ls` from line 35.

controllers/k8s_scan/container_image/deployment_handler.go
```diff
@@ -0,0 +1,130 @@
...
+
+type DeploymentHandler struct {
+	KubeClient             client.Client
+	Mondoo                 *v1alpha2.MondooAuditConfig
+	ContainerImageResolver mondoo.ContainerImageResolver
+	MondooOperatorConfig   *v1alpha2.MondooOperatorConfig
+}
+
+func (n *DeploymentHandler) Reconcile(ctx context.Context) (ctrl.Result, error) {
+	if !n.Mondoo.Spec.KubernetesResources.Enable && n.Mondoo.Spec.KubernetesResources.ContainerImageScanning {
```
czunker, line 42: I think the check for `!n.Mondoo.Spec.KubernetesResources.Enable` isn't needed.
When this is `false`, you will not reach this part of the code, because of https://github.com/mondoohq/mondoo-operator/pull/430/files#diff-6b5eb21c2966b28a0cf8b101ed7c6c9bb0eae9552c3690d47b96724cce30db05R52

(controllers/k8s_scan/deployment_handler.go - line 52)

imilchev: ah yes, you are right but that is not intended. The goal is that container image scanning should be disabled also if `KubernetesResources.Enable` is `false`. I will adjust the code

PR #440, opened 2022-07-12
📄 Docuemnt which types of workload are scanned ...

czunker:
... by the admission controller


docs/user-manual.md
```diff
@@ -132,6 +132,13 @@ And create/modify/delete a Pod in another window:
 kubectl delete pod -n mondoo-operator --selector control-plane=controller-manager
 ```
 
+### Scanned workload types
+
+Currently the admission controller scans workloads of these type:
+- Pods
+- Deployments
+- DaemonSets
```
imilchev, line 140: this is far from complete. Check [here](https://github.com/mondoohq/mondoo-operator/blob/main/controllers/admission/webhook-manifests.yaml). Also it is missing the details about not scanning pods that are children of DaemonSets, for example.

Having some short documentation here is good but I think we need more extensive documentation on our docs site as this is where most of the users are going

imilchev (changes requested):

docs/user-manual.md
czunker: Both done.

joelddiaz (approved):

czunker:
Merge is blocked by #446

PR #444, opened 2022-07-12
v0.5.2

joelddiaz:

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=mondoo-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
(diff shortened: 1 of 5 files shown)


bundle.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=mondoo-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.21.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.22.1
```
imilchev, line 9: is this change intended?

imilchev (approved):

bundle.Dockerfile
joelddiaz: > is this change intended?

According to https://github.com/mondoohq/mondoo-operator/pull/426 , we have already moved to 1.22

tas50:
@joelddiaz This has some new features in it. It should be 0.6.0

PR #446, opened 2022-07-13
🐛 Remove explicit version check and install latest release

czunker:
This prevents problems during the release process.
`git tag --list` will already include the tag we want to release,
and we would try an update from the new tag to the new tag.
Which isn't not yet released...
This fails because no release can be found on GitHub.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ VERSION ?= sha256-$(shell git rev-parse HEAD).sig
 # The latest url is just a HTTP redirect to an actual tag.
 # The actual tag is the last part of the location header
 # i.e.: location: https://github.com/mondoohq/mondoo-operator/releases/tag/v0.5.1
-PREVIOUS_RELEASE ?= $(shell curl -sI https://github.com/mondoohq/mondoo-operator/releases/latest/ | grep "location:" | awk -F"/" '{ print $NF }')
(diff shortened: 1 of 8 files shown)


imilchev (approved):

czunker:
Turns out, the GH actions ubuntu images do not have `awk` installed.

Makefile
```diff
@@ -5,7 +5,10 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= sha256-$(shell git rev-parse HEAD).sig
 
-PREVIOUS_RELEASE ?= $(shell git tag --list 'v*' | tail -1)
+# The latest url is just a HTTP redirect to an actual tag.
+# The actual tag is the last part of the location header
+# i.e.: location: https://github.com/mondoohq/mondoo-operator/releases/tag/v0.5.1
+PREVIOUS_RELEASE ?= $(shell $(curl -sI https://github.com/mondoohq/mondoo-operator/releases/latest/ | grep "location:"  | awk -F"/" '{ print $NF }'))
```
joelddiaz, line 11: just use the awesome power of awk and let awk do the search through the text for you
```suggestion
PREVIOUS_RELEASE ?= $(shell $(curl -sI https://github.com/mondoohq/mondoo-operator/releases/latest/ | awk -F"/" '/^location:/{ print $NF }'))
```

czunker: I will give this one a try in another PR. Also installing `awk` didn't help.
As this is blocking other things, I remove the explicit version check for now.

PR #454, opened 2022-07-19
🧹 Revised resources requests for operator and webhook

imilchev:
I noticed on my cluster that we are actually under-provisioning the operator and webhook. 

image
 


czunker (approved):
LGTM

PR #457, opened 2022-07-20
🧹 Update upgrade docs in prep for 1.0 behavior

joelddiaz:


imilchev (approved):
Sweet!

imilchev:
Perhaps we also want to remove the "early-access" stuff from the `README.md`. If we go for v1, the it's no longer early access I guess

PR #459, opened 2022-07-21
create separate default Node resource requirements

joelddiaz:
Having a 400m CPU request for the Node scanning Jobs is excessive for an hourly node scan where the completion time is not time sensitive.

Rather than have the Node resource defaults be shared with the singleton ScanAPI Deployment, allow for the Node default resource requirements to be different. Lower the CPU requests for the Node Jobs down to 50m so that scheduling a node scan doesn't request such a high amount of CPU on each node.

### DIFF
--- a/pkg/utils/k8s/resources_requirements.go
+++ b/pkg/utils/k8s/resources_requirements.go
@@ -21,7 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
-// defaultMondooClientResources for Mondoo Client container
+// DefaultMondooClientResources for Mondoo Client container
 var DefaultMondooClientResources corev1.ResourceRequirements = corev1.ResourceRequirements{
 	Limits: corev1.ResourceList{
 		corev1.ResourceMemory: resource.MustParse("400M"),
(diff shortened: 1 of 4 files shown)


pkg/utils/k8s/resources_requirements.go
```diff
@@ -44,3 +44,27 @@ func ResourcesRequirementsWithDefaults(m corev1.ResourceRequirements) corev1.Res
 	// Default values for Mondoo resources requirements.
 	return DefaultMondooClientResources
 }
+
+// DefaultNodeScanningResources for Mondoo Client container when scanning nodes
+var DefaultNodeScanningResources corev1.ResourceRequirements = corev1.ResourceRequirements{
+	Limits: corev1.ResourceList{
+		corev1.ResourceMemory: resource.MustParse("400M"),
```
imilchev, line 51: I think the memory is way too high for this. Verify what is the actual usage. I think it's not more than 100M. Probably a request of 60M and a limit 100M is suffcieint

pkg/utils/k8s/resources_requirements.go
```diff
@@ -44,3 +44,27 @@ func ResourcesRequirementsWithDefaults(m corev1.ResourceRequirements) corev1.Res
...
+		corev1.ResourceCPU:    resource.MustParse("200m"),
+	},
+
+	Requests: corev1.ResourceList{
+		corev1.ResourceMemory: resource.MustParse("180M"),
+		corev1.ResourceCPU:    resource.MustParse("50m"),
```
imilchev, line 57: Now that I think about this, while you are at it can you limit the resource requests for CPU for the k8s-scan and image scan cronjobs to 50m as well. They don't really do much apart from waiting for the scan API to finish it's task

imilchev (approved):

PR #460, opened 2022-07-22
🧹 no more 'control-plane' labeling

joelddiaz:
Using the default label of 'control-plane' that operator-sdk gives us means that the ValidatingWebhook when installed on Azure Kubernetes Service will ignore Resource activity in the mondoo-operator Namespace.

Migrate away from the 'control-plane' label use since it really doesn't make sense as mondoo-operator isn't a "control-plane"-type of workload.

Update the ValidatingWebhook to include an AKS-specific label so that the webhook does in fact get notified about resources even if they are in a Namespace with the 'control-plane' label.


config/crd/bases/k8s.mondoo.com_managedmondooauditconfigs.yaml
```diff
@@ -0,0 +1,93 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.9.0
+  creationTimestamp: null
+  name: managedmondooauditconfigs.k8s.mondoo.com
```
czunker, line 8: I think, this should be part of another PR.

config/manager/manager.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
-    control-plane: controller-manager
+    app: mondoo-operator
```
czunker, line 5: As we are chanign them, could we use the recommended labels from the docs?
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

controllers/admission/deployment_handler.go
```diff
@@ -334,6 +344,14 @@ func (n *DeploymentHandler) prepareValidatingWebhook(ctx context.Context, vwc *w
 		}
 	}
 
+	// For AKS the ValidatingWebhookConfiguration would normally not be notified about resources
+	// in Namespaces with the label key 'control-plane'. Adding the label "admissions.enforcer/disabled": "true"
+	// allows the Mondoo Operator webhook to be notified about these resources.
+	if vwc.Labels == nil {
+		vwc.Labels = map[string]string{}
+	}
+	vwc.Labels["admissions.enforcer/disabled"] = "true"
```
czunker, line 353: Now, that we have multiple labels for the webhook, perhaps it makes sense to use something similar to this:
https://github.com/mondoohq/mondoo-operator/blob/01ca21589c6444c9a788418fd6e3b8041ab76512/controllers/scanapi/resources.go#L275

controllers/admission/resources.go
```diff
@@ -16,8 +16,8 @@ import (
 )
 
 const (
-	WebhookLabelKey   = "control-plane"
-	WebhookLabelValue = "webhook-manager"
+	WebhookLabelKey   = "app"
+	WebhookLabelValue = "mondoo-operator"
```
czunker, line 20: See the comment for this line:
https://github.com/mondoohq/mondoo-operator/pull/460/files#diff-488f53c4545682df56fe8232f11bfedcc38fea937fff724c03d0240b6d6a4877R353

tests/framework/installer/installer.go
```diff
@@ -91,7 +91,12 @@ func (i *MondooInstaller) InstallOperator() error {
 		return err
 	}
 
-	if !i.K8sHelper.IsPodReady("control-plane=controller-manager", i.Settings.Namespace) {
+	watchLabel := "app=mondoo-operator"
+	// FIXME: remove this once we have migrated away from the old labeling scheme
+	if i.Settings.installRelease {
+		watchLabel = "control-plane=controller-manager"
+	}
+	if !i.K8sHelper.IsPodReady(watchLabel, i.Settings.Namespace) {
```
czunker, line 99: A comment like below, would be nice above this line: `// FIXME: remove up to here`

At least I will not remember.

czunker:
@imilchev fixed the spell checking in this PR: https://github.com/mondoohq/mondoo-operator/pull/461

config/manager/manager.yaml
joelddiaz: app.kubernetes.io/name = mondoo / mondoo-operator seems good to me

controllers/admission/deployment_handler.go
joelddiaz: The labels are actually distinct. One label is for the Deployment, and one label is for the ValidatingWebhookConfiguration. Either way, I did update to have a function that gives back the labels for the Deployment, and made the label values private to the package so that the function is the "official" way to get the list of labels.

config/crd/bases/k8s.mondoo.com_managedmondooauditconfigs.yaml
```diff
@@ -0,0 +1,93 @@
+---
```
czunker, line 1: You removed this file with one of the commits, but it sneaked back in.

config/manager/manager.yaml
```diff
@@ -11,16 +11,16 @@ metadata:
   name: controller-manager
   namespace: system
   labels:
-    control-plane: controller-manager
+    app.kubernetes.io/name: mondoo-operator
```
imilchev, line 14: This label is used by helm as well. Right now if you try to create a Helm chart with `helmify` it will create an invalid YAML because there are no labels for the Deployment. This is because internally it deletes `app.kubernete.io/name` since helm is also using that label. You can see the implementation they have [here](https://github.com/arttor/helmify/blob/eecc5e6fe220096d19e5a1b28ff4160e7cb7899d/pkg/processor/meta.go#L29)

joelddiaz: Can we do something like 'yq delete' the labels before passing it to kustomize/helmify?

czunker: But then, we would have different labels depending on the installation method, aren't we?

czunker: Or, we have to add it back later. Very strange behavior.

joelddiaz: I was thinking that helm would add them back so there wouldn't be a conflict...

imilchev: helm is adding that exact same label back, yes. However, I looked at 2 other operators and their helm charts :
- Rook
- prometheus-operator
They both have more than just `app.kubernetes.io/name` as a label. Feels like it is a common thing to have more than just that label on your deployments. That would also explain why `helmify` is implemented such that it cannot handle resources with just `app.kubernetes.io/name`.

I would prefer to avoid using `yq` to do custom stuff on top of the custom stuff that `helmify` does. I don't think having an extra label would hurt anyone 😆

joelddiaz: I'm not concerned about extra labels at all. Just wondering if there is a general expectation that `app.kubernetes.io/name` will be set for our resources, but then we don't have that label unless you happened to deploy with Helm. And looking at helmify, it looks like this functionality was added just 18 days ago :( https://github.com/arttor/helmify/pull/47 .

czunker: Sounds to me like Ivans PR is the correct way, but who knows when this will be merged. So, perhaps until then we use `app` and `app.kubernetes.io/name`?

imilchev:
I removed the extra labels for the deployment as `helmify` was fixed and we no longer need them

czunker (approved):
LGTM

PR #467, opened 2022-07-26
0.8.0

imilchev:

### DIFF
--- a/charts/mondoo-operator/Chart.yaml
+++ b/charts/mondoo-operator/Chart.yaml
@@ -13,9 +13,9 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.7.2
+version: 0.8.0
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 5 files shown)


imilchev:
@czunker since the upgrade integration tests are failing because of the label change, should we disable them for now? I am not sure how exactly they work. I also wonder why they succeeded in #460 but failed on `main`. Any ideas?

czunker:
> @czunker since the upgrade integration tests are failing because of the label change, should we disable them for now? I am not sure how exactly they work. I also wonder why they succeeded in #460 but failed on `main`. Any ideas?

#460 also changed the Upgrade Test to delete and then install again. So, this should be fine.
From the outputs in this PR, it seems to me, that the first install didn't work. The test didn't try to update.

I'll re-try. Perhaps minikube didn't come up or something like this.

czunker:
> > @czunker since the upgrade integration tests are failing because of the label change, should we disable them for now? I am not sure how exactly they work. I also wonder why they succeeded in #460 but failed on `main`. Any ideas?
> 
> #460 also changed the Upgrade Test to delete and then install again. So, this should be fine. From the outputs in this PR, it seems to me, that the first install didn't work. The test didn't try to update.
> 
> I'll re-try. Perhaps minikube didn't come up or something like this.

Re-try didn't help, but this passed: https://github.com/mondoohq/mondoo-operator/pull/468

PR #470, opened 2022-07-26
🐛 follow the platformid pattern

joelddiaz:


chris-rock (commented):
This should be the cluster id. Using the node id instead of the node name is probably better because users can then rename nodes

chris-rock (approved):
@joelddiaz Great improvement

PR #473, opened 2022-07-27
⭐ Allow enabling pod discovery with an env var

imilchev:
Allow enabling pod discovery by setting `FEATURE_DISCOVER_PODS` env var. If you'd like to test this you need to do the following:

1. Make sure minikube is running `minikube start`
2. In the `mondoo-operator` root directory run the following:
 ```bash
make generate-manifests
make load-minikube
kubectl apply -f mondoo-operator-manifests.yaml
``` 
3. Edit the mondoo-operator deployment to add the env var
```bash
kubectl edit -n mondoo-operator deployment mondoo-operator-controller-manager
```

It should look like this:
```yaml
....
containers:
  - args:
    - operator
    - --health-probe-bind-address=:8081
    - --metrics-bind-address=:8080
    - --leader-elect
    command:
    - /mondoo-operator
    env: # <-- this is the important bit
    - name: FEATURE_DISCOVER_PODS
      value: "1"
....
```

4. Apply the following `MondooOperatorConfig` config:
```yaml
kind: MondooOperatorConfig
metadata:
  name: mondoo-operator-config
spec:
  skipContainerResolution: true
```
5. Create a Mondoo space and upload policies that are targeting `k8s.pod` assets
6. Create a k8s integration in the Mondoo space and copy ONLY the command that creates the token
7. Apply the following `MondooAuditConfig`:
```yaml
kind: MondooAuditConfig
metadata:
  name: mondoo-client
  namespace: mondoo-operator
spec:
  mondooCredsSecretRef:
    name: mondoo-client
  mondooTokenSecretRef:
    name: mondoo-token
  scanner:
    image:
      name: docker.io/mondoo/client
      tag: edge-6.9.0-7
  kubernetesResources:
    enable: true
```

Within a minute the pods should show up in your Mondoo space

### DIFF
--- a/pkg/mondooclient/client.go
+++ b/pkg/mondooclient/client.go
@@ -24,6 +24,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"os"
 	"time"
 
 	"go.mondoo.com/mondoo-operator/pkg/constants"
@@ -37,6 +38,13 @@ const (
 	defaultTLSHandshakeTimeout = 10 * time.Second
 	maxIdleConnections         = 100
 )
+
+var enablePodDiscovery bool
+
+func init() {
+	_, ok := os.LookupEnv("SCAN_DISCOVER_PODS")
+	enablePodDiscovery = ok
+}
 
 //go:generate ./../../bin/mockgen -source=./client.go -destination=./mock/client_generated.go -package=mock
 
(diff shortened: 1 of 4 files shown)


pkg/mondooclient/client.go
```diff
@@ -38,6 +39,13 @@ const (
 	maxIdleConnections         = 100
 )
 
+var enablePodDiscovery bool
+
+func init() {
+	_, ok := os.LookupEnv("SCAN_DISCOVER_PODS")
```
joelddiaz, line 45: Do we need some way to pass down this environment variable from the mondoo-operator Deployment to the generated CronJob definitions that will instantiate their own MondooClient?

imilchev: Yes, we do. I just added the var to the client but forgot that it's called from the `CronJob` and not from the operator. Will update that once we can actually test it

joelddiaz (approved):

czunker (approved):
Works! Nice! :tada:

chris-rock:
created https://github.com/mondoohq/mondoo-operator/issues/474 to track the discussion around the `MondooOperatorConfig` flag

PR #476, opened 2022-07-28
⭐ Create edge integration tests workflow

imilchev:
- [x] split out integration tests into a re-usable workflow
- [x] fix current tests so they use the new workflow
- [x] add the possibility to override the Mondoo client tag via env var
- [x] create a new workflow that can be triggered manually by specifying a specific version of the Mondoo client

### DIFF
--- a/.github/workflows/edge-integration-tests.yaml
+++ b/.github/workflows/edge-integration-tests.yaml
@@ -19,4 +19,5 @@ jobs:
 
       - uses: ./.github/workflows/integration-tests.yaml
         with:
-          mondooClientImageTag: ${{ github.event.inputs.mondooClientImageTag }} 
+          mondooClientImageTag: ${{ github.event.inputs.mondooClientImageTag }}
+          secrets: inherit
--- a/.github/workflows/edge-integration-tests.yaml
+++ b/.github/workflows/edge-integration-tests.yaml
@@ -1,0 +1,22 @@
+name: Edge integration tests
+on:
+  workflow_dispatch:
+    inputs:
+      mondooClientImageTag:
+        description: "The Mondoo client image tag to be used for the integration tests"
+        required: true
+        type: string
+
+jobs:
+  integration_tests:
+    runs-on: ubuntu-latest
+    name: Start edge integration tests
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - uses: ./.github/workflows/integration-tests.yaml
+        with:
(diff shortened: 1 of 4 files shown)


imilchev:
I was hoping that I will be able to test this while it is still in a branch but I guess I can only see the new workflow once it gets merged into `main`. Will need to get this merged before I can see if it actually works.

tests/framework/utils/audit_config.go
```diff
@@ -26,6 +42,14 @@ func DefaultAuditConfigMinimal(ns string, workloads, nodes, admission bool) mond
 			Admission:            mondoov2.Admission{Enable: admission},
 		},
 	}
+
+	if mondooClientImageTag != "" {
+		auditConfig.Spec.Scanner.Image.Name = mondoo.MondooClientImage
```
joelddiaz, line 47: why setting the image name if only the tag was provided?

.github/workflows/edge-integration-tests.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  integration_tests:
+    runs-on: ubuntu-latest
+    name: Start edge integration tests
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
joelddiaz, line 18: maybe this is just my unfamiliarity with github actions, but the very first step in the github/workflows/integration-tests.yaml is precisely this step. why is it necessary here?

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,69 @@
...
+        with:
+          memory: 4000m
+          kubernetes-version: ${{ matrix.k8s-version }}
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "${{ env.golang-version }}"
+
+      - name: Store creds
+        run: |
+          echo ${{ secrets.MONDOO_CLIENT }} | base64 -d > creds.json
```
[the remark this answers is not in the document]
imilchev: Pre-pulling wasn't working until now either. It was pulling an image that we are not using. That's why I removed it

tests/framework/utils/audit_config.go
imilchev: I was testing some stuff. Indeed, this line can be removed

.github/workflows/edge-integration-tests.yaml
imilchev: Maybe it's not needed. I have no way of testing this right now so will just have to merge it and see what happens once it is started

chris-rock (approved):
Thank you @imilchev

PR #477, opened 2022-07-29
🐛 Fix edge integration workflow

imilchev:
Fix setting the secrets for the integration tests and print the inputs so we can more easily debug if needed

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -35,6 +35,10 @@ jobs:
         with:
           go-version: "${{ env.golang-version }}"
 
+      # Makes it easier to see what was the input for this workflow in case we need to debug.
+      - name: Print workflow inputs
+        run: echo "${{ toJSON(github.event.inputs) }}"
+
       - name: Store creds
         run: |
           echo ${{ secrets.MONDOO_CLIENT }} | base64 -d > creds.json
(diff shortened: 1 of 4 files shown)


.github/workflows/integration-tests.yaml
```diff
@@ -35,6 +35,10 @@ jobs:
         with:
           go-version: "${{ env.golang-version }}"
 
+      # Makes it easier to see what was the input for this workflow in case we need to debug.
+      - name: Print workflow inputs
+        run: echo "${{ toJSON(github.event.inputs) }}"
```
czunker, line 40: Do you have an example of what this contains?

We should be sure, that it does not contain sensitive data.

imilchev: it contains the block that is defined above as inputs. Currently, it's just the Mondoo client tag. I want to make sure we actually print the version otherwise it's impossible to find it... seems like github doesn't show nicely what input you used for a workflow...

imilchev: It looks like this when it's set:
```json
{
  mondooClientImageTag: edge-6.9.0-7
}
```

czunker: :+1: Thanks.

czunker (approved):
LGTM

PR #480, opened 2022-08-01
⭐ Add timeout parameter for scan cronjobs

imilchev:
Adds a timeout parameter for the scan `CronJobs`. The goal is to make sure that we don't keep stacking running scan `CronJobs`. Over the weekend I ran into the issue that the Mondoo client seems to get stuck occasionally during a scan or it gets killed due to OOM. If this happens the pending scan `CronJobs` remain pending forever and new ones are being added on top. This eventually eats all the cluster resources and the follow-up jobs are just queued to be scheduled.

This PR does NOT fix the Mondoo client issue but it does make sure that the operator properly handles such edge cases and does not eat up the cluster resources.

### DIFF
--- a/cmd/mondoo-operator/k8s_scan/cmd.go
+++ b/cmd/mondoo-operator/k8s_scan/cmd.go
@@ -2,9 +2,11 @@ package k8s_scan
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"go.mondoo.com/mondoo-operator/pkg/mondooclient"
@@ -22,6 +24,7 @@ func init() {
 	tokenFilePath := Cmd.Flags().String("token-file-path", "", "Path to a file containing token to use when making scan requests.")
 	integrationMrn := Cmd.Flags().String("integration-mrn", "", "The Mondoo integration MRN to label scanned items with if the MondooAuditConfig is configured with Mondoo integration.")
 	scanContainerImages := Cmd.Flags().Bool("scan-container-images", false, "A value indicating whether to scan container images.")
+	timeout := Cmd.Flags().Int64("timeout", 0, "The timeout in minutes for the scan request.")
 
 	Cmd.RunE = func(cmd *cobra.Command, args []string) error {
 		log.SetLogger(zap.New())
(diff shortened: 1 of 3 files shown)


cmd/mondoo-operator/k8s_scan/cmd.go
```diff
@@ -33,10 +36,17 @@ func init() {
...
 			return fmt.Errorf("--token-file-path must be provided")
 		}
+		if *timeout <= 0 {
+			return fmt.Errorf("--timeout must be greater than 0")
+		}
 
 		tokenBytes, err := ioutil.ReadFile(*tokenFilePath)
 		if err != nil {
-			logger.Error(err, "failed to read in file with token content")
+			if errors.Is(err, context.DeadlineExceeded) {
```
czunker, line 45: Does this error really happen here? I would have expected it on line 62, the actual scan.

imilchev: I didn't commit my latest changes... 😕

czunker (commented):
Not 100% sure, but I think this will not prevent CronJobs from stacking up.

When the timeout is reached, the Job will stop with an error, and because of the RestartPolicy, will be restarted: https://github.com/mondoohq/mondoo-operator/blob/main/controllers/k8s_scan/resources.go#L66

imilchev:
Jobs have backoff limit, which will eventually make sure that the pod is not restarted and will just fail. It will take a while but eventually it will result in a failed job and a terminated pod. The default value for the backoff limit is 6, which means the job will be restarted 6 times before it is marked as fail. I guess a value of 1 would be good enough for our scanning?

imilchev:
I figured another way to fix it:
```
// Specifies how to treat concurrent executions of a Job.
// Valid values are:
// - "Allow" (default): allows CronJobs to run concurrently;
// - "Forbid": forbids concurrent runs, skipping next run if previous run hasn't finished yet;
// - "Replace": cancels currently running job and replaces it with a new one
```

I set the `ConcurrencyPolicy` to `Forbid` so it means that it will skip the job if the old one is still running. Still the timeout addition is nice so we make sure jobs don't hang forever

czunker:
> ConcurrencyPolicy

Nice!

czunker (approved):

PR #482, opened 2022-08-02
⭐ Allow specifying extra env vars for scanner

imilchev:


controllers/scanapi/resources.go
```diff
@@ -255,6 +255,9 @@ func ScanApiDeployment(ns, image string, m v1alpha2.MondooAuditConfig, privateIm
 		})
 	}
 
+	// Merge the operator env for the scanner with the one provided in the MondooAuditConfig
+	scanApiDeployment.Spec.Template.Spec.Containers[0].Env = k8s.MergeEnv(scanApiDeployment.Spec.Template.Spec.Containers[0].Env, m.Spec.Scanner.Env)
```
czunker, line 259: With this ordering the user can overwrite env vars we set in resource.go with values specified in the MondooAuditConfig. Is this intended?

imilchev: Yes, it is documented in the function that does the env merge and it is also documented on `spec.scanner.env`. It is intended behaviour.  When you want to debug or try-out certain things, you are expected to know what you are doing. We cannot always force the values from the operator to have precedence as that kind of defeats the purpose of the whole thing we are doing here.

czunker (approved):

PR #484, opened 2022-08-04
🧹 Print full response on k8s scan error

imilchev:

### DIFF
--- a/cmd/mondoo-operator/k8s_scan/cmd.go
+++ b/cmd/mondoo-operator/k8s_scan/cmd.go
@@ -69,7 +69,7 @@ func init() {
 		if res.Ok {
 			logger.Info("Kubernetes resources scan successful", "worst score", res.WorstScore.Value)
 		} else {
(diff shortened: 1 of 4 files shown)


cmd/mondoo-operator/k8s_scan/cmd.go
```diff
@@ -69,7 +69,7 @@ func init() {
 		if res.Ok {
 			logger.Info("Kubernetes resources scan successful", "worst score", res.WorstScore.Value)
 		} else {
-			err := fmt.Errorf("scan API returned not OK. %v", res.WorstScore)
+			err := fmt.Errorf("scan API returned not OK. %+v", res)
```
czunker, line 72: With this, I think the comment in line 68 can be removed.

imilchev: It is there because my plan was to be able to print the whole report in the cronjob. Right now it still shows very minimal output. We postponed printing the report until the reporter becomes a public package

czunker (approved):

PR #489, opened 2022-08-11
🧹 Update go mod and build caching

imilchev:
Updated the go mod and build caching according to the official docs (see [here](https://github.com/actions/cache/blob/main/examples.md#go---modules))

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -48,28 +48,17 @@ jobs:
         with:
           go-version: "${{ env.golang-version }}"
 
-      # Get values for cache paths to be used in later steps
-      - id: go-cache-paths
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yaml
```diff
@@ -48,28 +48,17 @@ jobs:
         with:
           go-version: "${{ env.golang-version }}"
 
-      # Get values for cache paths to be used in later steps
-      - id: go-cache-paths
-        run: |
-          echo "::set-output name=go-build::$(go env GOCACHE)"
-          echo "::set-output name=go-mod::$(go env GOMODCACHE)"
-
       - name: Checkout
```
czunker, line 51: This can also be removed. e already did a checkout on line 44.

imilchev: done

czunker (approved):

PR #490, opened 2022-08-12
⭐ Upgrade to Go 1.19

imilchev:
Since we are upgrading the Mondoo client to Go 1.19 it would be good to have the operator upgraded as well

### DIFF
--- a/.github/env
+++ b/.github/env
@@ -1,2 +1,2 @@
-golang-version=1.18
+golang-version=1.19
 operator-sdk-version=v1.21.0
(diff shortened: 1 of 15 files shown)


.github/env
```diff
@@ -1,2 +1,2 @@
-golang-version=1.18
+golang-version=1.19
 operator-sdk-version=v1.21.0
```
czunker, line 2: That's the only thing I'm a bit worried about.

operator-sdk 1.22 started to support Go 1.18. No support for 1.19.

czunker (approved):
Let's give it a try.

.github/env
imilchev: i did run the manifest generation step and the crd generation step and it seems to just work. I think it should be fine

PR #495, opened 2022-08-18
⭐️ Upgrade k8s distros in integration tests

imilchev:


.github/workflows/integration-tests.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
     
     strategy:
       matrix:
-        k8s-version: [v1.22.12, v1.23.9, v1.24.3]
+        k8s-version: [v1.22.13, v1.23.10, v1.24.4]
```
czunker, line 23: Seems like k3s is lagging behind:
image

Couldn't find releases for the new versions: https://github.com/k3s-io/k3s/releases

Perhaps, we have to split them up?

imilchev: Nah, I think we can just wait until they add support for those versions in k3s as well. They are quite quick usually

czunker (approved):

PR #496, opened 2022-08-18
🧹 Add EKS terraform templates

czunker:


imilchev:
Is the goal to automate this or is it going to be done in a separate PR? It is a bit strange that we have a terraform template in the repo for an operator. If this is going to be used just for the sake of testing I would put it either in the tests folder or under .github

PR #497, opened 2022-08-22
Use check-spelling/check-spelling@v0.0.20

jsoref:
Refreshes the workflow based on
https://github.com/check-spelling/spell-check-this/blob/744c66e2140fd8acaf5388efd0db3727d010d6e9/.github/workflows/spelling.yml

See the [release notes](https://github.com/check-spelling/check-spelling/releases/tag/v0.0.20) for a list of improvements/changes.

Note that `extra_dictionaries` is a 0.0.20 feature, so with this update, those values will be used (in v0.0.19, you just get warnings about unrecognized parameters).


joelddiaz (approved):

.github/actions/spelling/allow.txt
```diff
@@ -0,0 +1,10 @@
+Backend
+backend
+Cron
```
tas50, line 3: @jsoref What's the reasoning for moving some of the expect words to allow, but not all of them?

jsoref: Since I'm upgrading from one engine to another and they have a slightly different belief about which words are uppercase/lowercase, I don't want to deal w/ them. Words in `allow.txt` are treated as being in the dictionary, if they aren't present, that's fine, if they are, that's also fine. If I use `expect.txt` and the old version expects a different case, it'll scream (or vice versa). -- I try to be conservative about adding words to the dictionary.

If you know that a term is from your project's perspective a "word" whether or not it's currently in use, you could (should?) consider moving it to `allow.txt`.

PR #511, opened 2022-08-25
🧹 check for required APIs at startup

joelddiaz:
Rather than fail later while the operator is up and running, just check
whether some of the resources we depend on to operator properly exist
in the cluster.

For the first pass, just check that Jobs exist (to prove that our
checks actually work) and whether CronJobs exist (to catch the case of
an EKS cluster running k8s 1.20 which lacks CronJobs v1 (but happens to
have CronJobs v1beta1)).

### DIFF
--- a/cmd/mondoo-operator/operator/cmd.go
+++ b/cmd/mondoo-operator/operator/cmd.go
@@ -1,12 +1,16 @@
 package operator
 
 import (
+	"fmt"
+
+	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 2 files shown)


cmd/mondoo-operator/operator/cmd.go
```diff
@@ -78,6 +82,17 @@ func init() {
...
+			return err
+		}
+		if !passed {
+			msg := "required API(s) not found"
+			setupLog.Info(msg)
+			return fmt.Errorf(msg)
+		}
+
 		// The API group "config.openshift.io" should be unique to an OpenShift cluster
 		isOpenShift, err := k8s.VerifyAPI("config.openshift.io", "v1", setupLog)
```
czunker, line 97: Couldn't we just re-use this for the checks?

joelddiaz: I initially tried to use the VerifyAPI check, but it was passing when I was expecting a failure on 1.20.
For the k8s 1.20 situation, the batch/v1 API does exist, but only Jobs are defined there. CronJobs was available under batch/v1beta1 (which we of course don't use in our repo).

czunker (approved):

PR #514, opened 2022-08-26
🧹 extra dependabot ignores

joelddiaz:
As seen from some of the releases of OperatorSDK, there are some
packages whose version's we follow based on OperatorSDK
instructions/guidance.

https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.17.0/

Mark ginkgo, gomega, k8s.io/apimachinery, and k8s.io/client-go as packages
to ignore.


imilchev:
I think we don't even use ginkgo anymore. I believe all tests have been migrated so perhaps we can even drop that dependency

czunker (approved):

PR #516, opened 2022-08-29
🧹 Add EKS test workflow

czunker:


czunker:
Looks like I need to merge this branch before I can manually test it.

.github/workflows/release-tests.yaml
```diff
@@ -0,0 +1,110 @@
...
+        required: true
+        type: string
+        default: nightly
+        description: The image tag to use for the mondoo operator image
+      awsRegion:
+        required: true
+        type: string
+        default: us-east-2
+        description: AWS region where we create the EKS cluster
+      AWS_ACCESS_KEY_ID:
```
imilchev, line 20: why aren't all these AWS things secrets? I wouldn't expect that I need to be pasting AWS secrets in order to start the workflow

.github/workflows/release-tests.yaml
```diff
@@ -0,0 +1,110 @@
...
+        run: terraform apply -auto-approve ${{ matrix.k8s-distro }}.json
+        working-directory: ./github/terraform/${{ matrix.k8s-distro }}
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "${{ env.golang-version }}"
+
+      # Makes it easier to see what was the input for this workflow in case we need to debug.
+      - name: Print workflow inputs
+        run: echo "${{ toJSON(github.event.inputs) }}"
```
imilchev, line 81: with the current setup this is going to print the AWS secrets into the test log

czunker: Removed

.github/workflows/release-tests.yaml
czunker: On the first iteration, I'd like to start it manually. And also see it is really working.

When we agree on when to trigger this workflow automatically, we can add some secrets to GitHub to run it.

imilchev: Technically, you can start it manually and still have the secrets in Github. The `MONDOO_CLIENT` secret below works exactly like this. I wouldn't see why you want to explicitly provide the AWS creds for every workflow run.

If you still want to manually paste it then keep it as is. Is the `AWS_SESSION_TOKEN` needed? I thought that just the key and secret access key are sufficient...

imilchev (approved):

PR #519, opened 2022-08-29
Updates license to MPL2

scottford-io:
Signed-off-by: Scott Ford <scott@scottford.io>


imilchev:
I think this one can be closed now since we merged #532

PR #532, opened 2022-08-30
📝 Relicense as MPL 2.0

tas50:
Add the header from https://www.mozilla.org/en-US/MPL/headers/ and add
headers in many files that were missing.

Signed-off-by: Tim Smith <tsmith84@gmail.com>


pkg/utils/mondoo/condition.go
```diff
@@ -1,17 +1,9 @@
 /*
```
joelddiaz, line 1: I want to say a big chunk of this comes from https://github.com/openshift/cluster-operator/blob/master/pkg/controller/controller_utils.go . I could've sworn it came from somewhere else at the time, but my googling has failed me.

This should bye 2017 APL 2.0

imilchev (approved):
LGTM. If nobody has anything to add, let's merge this

joelddiaz:
> LGTM. If nobody has anything to add, let's merge this

Just the one file that comes from an APL2 source.

imilchev:
> > LGTM. If nobody has anything to add, let's merge this
> 
> Just the one file that comes from an APL2 source.

Not sure if that's the correct solution but I copy-pasted the license from the link that you provided. If that's correct, then we should be good to go

joelddiaz (approved):

PR #534, opened 2022-08-31
🧹 Fix versioning of the binaries when building containers

imilchev:


.github/workflows/publish.yaml
```diff
@@ -87,7 +83,7 @@ jobs:
             suffix=-${{ matrix.arch }},onlatest=true
 
       - name: Build binaries
-        run: VERSION=${{ steps.get_version.outputs.VERSION }} TARGET_OS=${{ matrix.os }} TARGET_ARCH=${{ matrix.arch }} make build
+        run: VERSION=${{ steps.meta.outputs.version }} TARGET_OS=${{ matrix.os }} TARGET_ARCH=${{ matrix.arch }} make build
```
czunker, line 86: Ok, from the registry I see Docker has a nightly tag. But I do not understand why Docker knows about this. Where is this nightly coming from?

imilchev: from that `meta` step. We have a "scheduled" run for that workflow that runs once a day. For scheduled events by default that step generates a "nightly" tag

czunker: Ah, thanks.

czunker (approved):

PR #535, opened 2022-08-31
🐛 Fix integration tests on EKS

czunker:


imilchev:
I suppose this fixes #536

czunker:
> I suppose this fixes #536

Yes, the one commit. I hope when I do not squash, that GitHub automatically closes it.

controllers/nodes/resources.go
```diff
@@ -21,6 +21,7 @@ import (
 const (
 	CronJobNameBase        = "-node-"
 	InventoryConfigMapBase = "-node-inventory-"
+	CronJobNameMaxLength   = 52
```
imilchev, line 24: It is actually the resource name max length in k8s. It's not cronjob-specific. It's just that we use it for the cronjobs because we add the node names to the cronjob name and in some cases the node names can be very long

imilchev (commented):
Are you going to rebase this PR on `main` or is that going to be done in another PR?

.github/workflows/release-tests.yaml
```diff
@@ -46,31 +46,49 @@ jobs:
     strategy:
       matrix:
         k8s-version: ["1.23"]
-        k8s-distro: ["aws"]
+        k8s-distro: ["eks"]
```
imilchev, line 49: I think you don't need this. You can list just the `k8s-version` like I Did for AKS.

.github/workflows/release-tests.yaml
```diff
@@ -46,31 +46,49 @@ jobs:
     strategy:
       matrix:
```
imilchev, line 47: It is important to add the `fail-fast: false` since if 1 of the k8s versions fails it will cancel all pending tests before the `terraform destroy`. That will leave dangling resources in AWS. I had that issue with AKS earlier today.

.github/workflows/release-tests.yaml
```diff
@@ -83,12 +101,22 @@ jobs:
       - name: Run integration tests
         run: VERSION=${{ env.MONDOO_OPERATOR_IMAGE_TAG }} K8S_DISTRO=${{ matrix.k8s-distro }} make test/integration/ci
         env:
-          KUBE_CONFIG: ./github/terraform/${{ matrix.k8s-distro }}/kubeconfig
+          AWS_ACCESS_KEY_ID: ${{ github.event.inputs.AWS_ACCESS_KEY_ID }}
```
imilchev, line 104: why are these needed for the integration tests?

.github/workflows/release-tests.yaml
```diff
@@ -83,12 +101,22 @@ jobs:
...
+          AWS_SECRET_ACCESS_KEY: ${{ github.event.inputs.AWS_SECRET_ACCESS_KEY }}
+          AWS_SESSION_TOKEN: ${{ github.event.inputs.AWS_SESSION_TOKEN }}
 
       - name: Clean up EKS terraform
-        run: terraform destroy
+        run: terraform destroy -auto-approve
+        env:
+          TF_VAR_test_name: ${{ env.MONDOO_OPERATOR_IMAGE_TAG }}
+          TF_VAR_kubernetes_version: ${{ matrix.k8s-version }}
+          AWS_ACCESS_KEY_ID: ${{ github.event.inputs.AWS_ACCESS_KEY_ID }}
```
imilchev, line 113: why not set these env vars at the job level instead of copy-pasting 5-6 lines for every step. Would make things easier to read

Makefile
```diff
@@ -109,7 +109,7 @@ test/ci: manifests generate fmt vet envtest gotestsum
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) --arch=amd64 use $(ENVTEST_K8S_VERSION) -p path)" $(GOTESTSUM) --junitfile unit-tests.xml -- $(UNIT_TEST_PACKAGES) -coverprofile cover.out
 
 # Integration tests are run synchronously to avoid race conditions
-ifeq ($(K8S_DISTRO),aws)
+ifeq ($(K8S_DISTRO),eks)
```
imilchev, line 112: I think on `main` it now says `aks`. Is there a way to do logical OR in this `ifeq`? I am a bit of a Makefile/bash noob

.github/workflows/release-tests.yaml
czunker: Done.

.github/workflows/release-tests.yaml
czunker: Thanks for the hint.

Makefile
czunker: I just looked it up. There is a construct you could use, but I do not think it is actually readable, so I used a more expressive solution.

controllers/nodes/resources.go
czunker: Moved that to a general package

czunker:
> Are you going to rebase this PR on `main` or is that going to be done in another PR?
> 
> Nevermind, I saw there's a conflict now

I re-based on main, but I kept two separate files. The input fields are so different, that I do not think it makes sense to combine them.

.github/workflows/release-tests.yaml
czunker: Deleted.

czunker:
> > > Are you going to rebase this PR on `main` or is that going to be done in another PR?
> > > Nevermind, I saw there's a conflict now
> > 
> > 
> > I re-based on main, but I kept two separate files. The input fields are so different, that I do not think it makes sense to combine them.
> 
> I think the inputs are identical. Why don't you arrange the AWS creds to be added to the repo and remove the secrets input. Then we have just the operator and client versions as input. I think we should have 1 workflow for all clouds and then we have 1 matrix job for each provider

Missed the point, that secrets are separate.

Changed the files accordingly.

.github/workflows/cloud-tests.yaml
```diff
@@ -116,10 +126,104 @@ jobs:
...
+    env:
+      TF_VAR_test_name: ${{ github.event.inputs.mondooOperatorImageTag }}
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      AWS_REGION: ${{ github.event.inputs.AWS_REGION }}
+
+    steps:
+      - name: hide values
+        run: |
+          echo "::add-mask::${{ github.event.inputs.AWS_ACCESS_KEY_ID }}"
```
imilchev, line 149: now that these are actual secrets I don't think this step is needed

.github/workflows/cloud-tests.yaml
```diff
@@ -12,6 +12,11 @@ on:
         type: string
         default: main
         description: The image tag to use for the mondoo operator image
+      AWS_REGION:
```
imilchev, line 15: why not just hard-code this into the workflow. I wouldn't see why you would want it to be different

.github/workflows/cloud-tests.yaml
czunker: Oh, they must have slipped back in during one of the rebases. Good catch. Thanks.

.github/workflows/cloud-tests.yaml
czunker: During testing, I created them in the EU, but agree.

.github/workflows/cloud-tests.yaml
```diff
@@ -116,10 +121,98 @@ jobs:
...
+        if: success() || failure()
+
+      - run: mv integration-tests.xml integration-tests-eks-${{ matrix.k8s-version }}.xml
+        if: success() || failure()
+
+      - name: Upload test results
+        uses: actions/upload-artifact@v3  # upload test results
+        if: success() || failure()        # run this step even if previous step failed
+        with:                             # upload a combined archive with unit and integration test results
+          name: test-results
```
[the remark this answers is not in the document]
czunker: Thanks. Fixed.

imilchev (approved):

PR #541, opened 2022-09-01
🧹 Cleanup Mondoo space after integration tests

czunker:

### DIFF
--- a/.github/workflows/cleanup-mondoo-space.yaml
+++ b/.github/workflows/cleanup-mondoo-space.yaml
@@ -1,0 +1,28 @@
+name: "Cleanup Mondoo space"
+on:
+  schedule:
+  - cron: "11 3 * * *" # 3:11am UTC
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    name: Cleanup Assets and Projects in Mondoo
+    steps:
+      - uses: actions/checkout@v2
+      - name: Store cleanup creds
+        run: |
+          echo ${{ secrets.MONDOO_CLIENT_EDITOR }} | base64 -d > creds_editor.json
+      - uses: mondoohq/actions/setup@main
+        if: success() || failure()
+        with:
+          service-account-credentials: ${{ secrets.MONDOO_CLIENT_EDITOR }}
+          args: status
+
+      - name: Cleanup Mondoo space "GitHub Actions Mondoo Operator"
+        if: success() || failure()
+        run: |
+          mondoo --config ${{ env.MONDOO_CONFIG_PATH }} status
+          ./scripts/cleanup_operator_assets.sh
+          ./scripts/cleanup_operator_space_cicd.sh
+        env:
+          MONDOO_CONFIG_PATH: ${{ format('{0}/{1}', github.workspace, 'creds_editor.json') }}
--- a/scripts/cleanup_operator_assets.sh
+++ b/scripts/cleanup_operator_assets.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+echo "#########################"
+echo " Deleteing Assets"
+echo "#########################"
+
 SPACE_MRN=$(jq '.space_mrn' -r  ./creds_editor.json)
 if [[ $SPACE_MRN == "" ]]
 then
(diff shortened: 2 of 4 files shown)


czunker:
As discussed, this might get easier when the Garbage Collection is implemented. Then we can run it as a schedules job to delete assets older than X days. Garbage Collection might also fix the permission problem. Otherwise we would need a different role for our service accout.

czunker:
@benr added a new Secret, so we can now delete the assets and projects.

We cannot use the delete filters @joelddiaz added because they aren't exposed via the API. So, I opted for a scheduled job that hopefully does not interfere with our tests. Otherwise, we would need to extend the API or write another client.

.github/workflows/cleanup-mondoo-space.yaml
```diff
@@ -0,0 +1,28 @@
...
+          service-account-credentials: ${{ secrets.MONDOO_CLIENT_EDITOR }}
+          args: status
+
+      - name: Cleanup Mondoo space "GitHub Actions Mondoo Operator"
+        if: success() || failure()
+        run: |
+          mondoo --config ${{ env.MONDOO_CONFIG_PATH }} status
+          ./scripts/cleanup_operator_assets.sh
+          ./scripts/cleanup_operator_space_cicd.sh
+        env:
```
imilchev, line 27: I think this is not needed. The path can be relative. It should also work if you just say `mondoo --config creds_editor.json status`

imilchev: ah now I see that your script uses that env var. Nevermind...

scripts/cleanup_operator_assets.sh
```diff
@@ -0,0 +1,89 @@
...
+        \"platformRuntime\":[],
+        \"labels\":[],
+        \"eol\":null,
+        \"reboot\":null,
+        \"exploitable\":null,
+        \"updated\":null,
+        \"orderBy\":
+        {
+          \"field\":\"LAST_UPDATED\",
+          \"direction\":\"DESC\"
```
imilchev, line 48: shouldn't this be `ASC`? If we do `DESC` I think we are deleting newest assets first

imilchev: can we delete only assets that are older than 1d? This will make sure that if we run this script in parallel with a test run, it won't interfere. We could also delete assets older than 1h but I thought 1d might be easier to implement

imilchev:
It almost feels like this could have been implemented in a much easier way if we write a small Go program that does just the cleanup

scripts/cleanup_operator_assets.sh
czunker: I can use ASC. But I think our API is currently not able to do searches like "last updated" before 1h or 1d.

But I could add a check not to delete all elements and keep the newest X assets. The same for CI/CD. Then the daily job wouldn't interfere with running tests.

czunker:
> It almost feels like this could have been implemented in a much easier way if we write a small Go program that does just the cleanup

For "easier" I'm not sure. More comfortable, definitely. I guess the go client will be the next iteration, but with UX and product involved. So it will take longer, just because more people are involved.

For now, I'd like to stick with the bash solution be it solves our current problem.

chris-rock:
We work an generated clients for the graphql api soon

scripts/cleanup_operator_assets.sh
czunker: Added a check to keep the newest items for fleet and ci/cd.

imilchev (approved):

joelddiaz:
> We cannot use the delete filters @joelddiaz added because they aren't exposed via the API. So, I opted for a scheduled job that hopefully does not interfere with our tests. Otherwise, we would need to extend the API or write another client.
> 
If I read that bash script correctly, it looks like it is performing a raw DeleteAssets() call, and that is fundamentally what the garbage collection does. Why wouldn't the new filtering parameters to that call be usable? It seems like using the `DateFilter` would allow deleting anything older than whatever date is provided to the call.

czunker:
@joelddiaz 
I tried to add the filter, but it didn't work. My understanding of the code was that the GQL exposes less than RPC.

PR #545, opened 2022-09-05
🧹 Scan our artifacts with Mondoo

czunker:

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -39,6 +39,12 @@ jobs:
     needs: [unit-tests]
     runs-on: ubuntu-latest
     name: Security tests
+    env:
+      # Use docker.io for Docker Hub if empty
+      REGISTRY: ghcr.io
+      # github.repository as <account>/<repo>
+      IMAGE_NAME: ${{ github.repository }}
+      RELEASE: ${{ github.ref_name }}
     steps:
       - uses: actions/checkout@v3
       - name: Import environment variables from file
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yaml
```diff
@@ -35,6 +35,42 @@ jobs:
         with:
           name: test-results
           path: unit-tests.xml
+  security-tests:
+    needs: [unit-tests]
+    runs-on: ubuntu-latest
+    name: Security tests
+    env:
+      # Use docker.io for Docker Hub if empty
+      REGISTRY: ghcr.io
```
imilchev, line 44: I think these are not needed

.github/workflows/tests.yaml
```diff
@@ -35,6 +35,42 @@ jobs:
...
+      - name: Scan Kubernetes Manifest with Mondoo
+        uses: mondoohq/actions/k8s-manifest@main
+        with:
+          service-account-credentials: ${{ secrets.MONDOO_CLIENT }}
+          path: "mondoo-operator-manifests.yaml"
+      - uses: actions/upload-artifact@v3  # upload test results
+        if: success() || failure()        # run this step even if previous step failed
+        with:
+          name: test-results
+          path: security-tests.xml
```
imilchev, line 73: what is generating this file?

.github/workflows/tests.yaml
```diff
@@ -35,6 +35,42 @@ jobs:
...
+      IMAGE_NAME: ${{ github.repository }}
+      RELEASE: ${{ github.ref_name }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Import environment variables from file
+        run: cat ".github/env" >> $GITHUB_ENV
+      - name: Scan AWS terraform with Mondoo
+        uses: mondoohq/actions/terraform@main
+        with:
+          service-account-credentials: ${{ secrets.MONDOO_CLIENT }}
```
imilchev, line 55: This will add the scans to the big space with the integration tests. It will be a bit difficult to find them but I guess that is the right place

.github/workflows/tests.yaml
czunker: One tab too much. That was Autopilot. :laughing:

.github/workflows/tests.yaml
czunker: I need something like this for the manifest otherwise, it looks strange. So, I opted for the same values as in the release step.

imilchev (approved):

PR #547, opened 2022-09-06
🚀 1.3.2

czunker:

### DIFF
--- a/charts/mondoo-operator/Chart.yaml
+++ b/charts/mondoo-operator/Chart.yaml
@@ -13,9 +13,9 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.3.1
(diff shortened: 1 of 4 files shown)


imilchev:
any reason for the feature bump? From what I was able to see we have only cleanup and test stuff since the previous release

czunker:
> any reason for the feature bump? From what I was able to see we have only cleanup and test stuff since the previous release

The AKS tests have a feature star.

imilchev:
> > any reason for the feature bump? From what I was able to see we have only cleanup and test stuff since the previous release
> 
> The AKS tests have a feature star.

true but they are something we care about, not the end-users

czunker:
> > > any reason for the feature bump? From what I was able to see we have only cleanup and test stuff since the previous release
> > 
> > 
> > The AKS tests have a feature star.
> 
> true but they are something we care about, not the end-users

Agree. Changed to 1.3.2. For the future we should think about, how we could separate internal from user-facing features in the PRs.

imilchev (approved):

PR #549, opened 2022-09-07
🧹 apply changes from using v4 kubebuilder

joelddiaz:
Went through the OperatorSDK tutorial twice. Once with the traditional/default v3 and once with --plugins="go/v4-alpha" as suggested in https://github.com/kubernetes-sigs/controller-runtime/issues/1657#issuecomment-1236901482 .

Did a recursive diff between the two directories, and these were the differences.

We already have pretty much moved to v4 thanks to the changes in https://github.com/mondoohq/mondoo-operator/pull/411 , but this PR just attempts to apply some of the additional changes that would be made to the files if we had initialized the repo with "--plugins="go/v4-alpha" originally.

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -9,8 +9,10 @@ namespace: mondoo-operator
 namePrefix: mondoo-operator-
 
 # Labels to add to all resources and selectors.
-#commonLabels:
-#  someName: someValue
+#labels:
+#- includeSelectors: true
+#  pairs:
+#    someName: someValue
 
 bases:
 - ../crd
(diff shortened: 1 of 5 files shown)


config/default/kustomization.yaml
```diff
@@ -47,32 +49,102 @@ patchesStrategicMerge:
...
-#    name: serving-cert # this name should match the one in certificate.yaml
-#  fieldref:
-#    fieldpath: metadata.namespace
-#- name: CERTIFICATE_NAME
-#  objref:
-#    kind: Certificate
-#    group: cert-manager.io
-#    version: v1
-#    name: serving-cert # this name should match the one in certificate.yaml
-#- name: SERVICE_NAMESPACE # namespace of the service
```
czunker, line 67: As this changes here, do we also need to change it here: https://github.com/mondoohq/mondoo-operator/blob/9fda2b0a7ca58dd9258a39e279941eabe4f7b71f/config/certmanager/certificate.yaml ?

Basically the same change as for `$(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)`

joelddiaz: Since we do our own version of supporting cert-manager, these are just boilerplate changes that OperatorSDK produces for us that we just simpoly don't use.
To be honest, these seem like bugs, but as much as possible I just wanted to keep our boilerplate (whether used or unused) in sync with what OperatorSDK produces.

czunker (approved):

PR #556, opened 2022-09-13
🧹 Enable dependabot PRs

### LINKED ISSUE #553: Cannot run integration tests for forks
Our integration tests fail for forked repos and dependabot branches because they do not have access to the repo secrets. For now the main issue is the dependabot PRs because we cannot merge them due to the failing tests.

More info about workarounds can be found in https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544 and [here](https://hugo.alliau.me/2021/05/04/migration-to-github-native-dependabot-solutions-for-auto-merge-and-action-secrets/#share-your-secrets-with-dependabot).

We use a similar approach for doing the test reporting for our integration tests. The reporting is separated into its own workflow which gets triggered once the integration tests finish. The reporting workflow then always runs from our repo and therefore has access to the repo and its secrets. We should try out and see what works for the integration tests.

czunker:


.github/workflows/integration-tests.yaml
```diff
@@ -17,6 +17,22 @@ jobs:
   integration-tests:
     runs-on: ubuntu-latest
     name: Integration tests
+    if: |
+      (
+        github.event_name == 'pull_request'
+        && (
+          github.actor == 'dependabot[bot]'
```
czunker, line 24: This list isn't that nice. Perhaps we could remove it and solve with the described approaches here: https://github.com/mondoohq/mondoo-operator/issues/553#issuecomment-1245338069

imilchev: Why do we need to check the actors explicitly? Can't we just check if we are running a fork?

czunker: Nice!
That makes it much more readable. Thanks

.github/workflows/tests.yaml
```diff
@@ -35,37 +40,13 @@ jobs:
...
-        run: make generate-manifests IMG='${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.RELEASE }}'
-      - name: Scan Kubernetes Manifest with Mondoo
-        uses: mondoohq/actions/k8s-manifest@main
-        with:
-          service-account-credentials: ${{ secrets.MONDOO_CLIENT }}
-          path: "mondoo-operator-manifests.yaml"
+    secrets: inherit
   integration-tests:
     needs: [unit-tests]
     uses: ./.github/workflows/integration-tests.yaml
```
imilchev, line 52: Do we know for sure this works? I think with this setup we will have the same issue with the secrets not being present

czunker: I'm not sure.

According to the docs, the `workflow_call` jobs have access to secrets. But I'm not sure what in this combination the `secrets: inherit` does. I couldn't find anything about this combination. In case this does not work, we can call the `integration-tests` and `security-tests` differently and try that.

For testing the secrets, I wanted to recreate this dependabot PR: https://github.com/mondoohq/mondoo-operator/pull/554

.github/workflows/integration-tests.yaml
```diff
@@ -18,22 +18,9 @@ jobs:
...
-          || github.actor == 'misterpantz'
-          || github.actor == 'scottford-io'
-          || github.actor == 'atomic111'
-          || github.actor == 'czunker'
-        )
-      )
-      || (contains(github.event.pull_request.labels.*.name, 'safe to test'))
-    
+      ! github.event.pull_request.head.repo.fork ||
+      (github.event.pull_request.head.repo.fork && github.actor == 'dependabot[bot]') ||
```
joelddiaz, line 22: I thought we had establlished that the dependabot PRs are not from forks? See https://github.com/mondoohq/mondoo-operator/pull/554

czunker: Changed. Thanks for the pointer.

czunker:
I changed this PR to only allow dependabot. We'll give the Settings described in #553 a try.

.github/workflows/tests.yaml
imilchev: I am almost certain that this won't change anything. If you use `secrets: inherit`, it means the child workflow is getting the secrets from the parent workflow. If the tests are running from a fork or for a dependabot branch, they have no access to the secrets. The integration tests won't have access either.

czunker: I switched to `pull_request_target` + checkout head. Now the jobs have access to the secrets.

But this also means we trust dependabot because the Jobs are now running with the code of the PR.

czunker:
So, next take on this problem.

According to the docs:
> The workflow started by the workflow_run event is able to access secrets and write tokens, even if the previous workflow was not.

.github/workflows/tests.yaml
```diff
@@ -11,9 +11,13 @@ on:
 jobs:
   unit-tests:
     runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]' || !github.event.pull_request_target.head.repo.fork
```
czunker, line 14: When we remove this line, even fork PRs should work. We could handle their approval with GitHub security settings as mentioned in https://github.com/mondoohq/mondoo-operator/issues/553#issuecomment-1245338069

joelddiaz: We would block the workflows from running via the approvals/tags on the PR then?

czunker: Yes.
@benr has activated the security settings mentioned in the comment. I understand we would see an approval button for dependabot PRs and fork PRs when this line is gone.

joelddiaz (approved):
can't wait for this one...let's see how this goes

PR #561, opened 2022-09-14
⭐️ Scan workloads on create/update

imilchev:
- [x] Implement resource monitor controller
- [x] Debounce events related to the same resource
- [x] Extend RBAC rules to allow the operator to read all workload types
- [x] Retrieve all `MondooAuditConfigs` with `kubernetesResources` enabled
- [x] Trigger a scan for the resources for all `MondooAuditConfigs` with `kubernetesResources` enabled
- [x] Unit tests


controllers/resource_monitor/scan_api_store/scan_api_store_test.go
```diff
@@ -0,0 +1,118 @@
...
+	url := utils.RandString(10)
+	token := utils.RandString(10)
+	integrationMrn := utils.RandString(10)
+	s.scanApiStore.mondooClientBuilder = func(opts mondooclient.ClientOptions) mondooclient.Client {
+		s.Equal(url, opts.ApiEndpoint)
+		s.Equal(token, opts.Token)
+		return s.mockMondooClient
+	}
+
+	for i := 0; i < 100; i++ {
```
czunker, line 74: :smile: 100 loops. Just to be sure ...

controllers/resource_monitor/scan_api_store/scan_api_store_test.go
```diff
@@ -0,0 +1,118 @@
...
+	integrationMrn := utils.RandString(10)
+	s.scanApiStore.mondooClientBuilder = func(opts mondooclient.ClientOptions) mondooclient.Client {
+		s.Equal(url, opts.ApiEndpoint)
+		s.Equal(token, opts.Token)
+		return s.mockMondooClient
+	}
+
+	s.scanApiStore.Add(url, token, integrationMrn)
+
+	url2 := url + "1"
```
czunker, line 97: Isn't that then "url1" ;)

pkg/mondooclient/client.go
```diff
@@ -306,6 +307,67 @@ func (s *mondooClient) ScanKubernetesResources(ctx context.Context, integrationM
...
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	if integrationMrn != "" {
```
czunker, line 336: We have this feature flag logic already somewhere else. Could we merge it into a function?

imilchev: I can extract it into a function. It's not a feature flag though. This is because it is possible to run the operator with an integration or without and we need to cover both cases

controllers/resource_monitor/scan_api_store/scan_api_store_test.go
imilchev: safety first 😆

pkg/mondooclient/client.go
czunker: I marked the wrong line. I meant L343 + L347. But I just checked the other code in the client.go. It is a bit different. Think now it wouldn't really make sense to extract it.

czunker (approved):

config/rbac/role.yaml
```diff
@@ -83,15 +103,6 @@ rules:
   verbs:
   - create
   - delete
-- apiGroups:
```
joelddiaz, line 86: Why did we drop the restricted list of access to secrets?

joelddiaz: Haven't we run across these kinds of cases before and landed on the side of "if the user is doing non-default things, then the user is responsible for updating the permissions the operator needs"?

If the default RBAC was written in a way where the "default/standard" MondooAuditConfig and it's default Secret names worked out of the box, and anything else would require updating RBAC, wouldn't that return us to how things have been previously developed?

PR #565, opened 2022-09-16
⭐️ k8s asset garbage collection

joelddiaz:
* set ManagedBy field on scanned/discovered assets

This depends on current features in edge that allow setting this field
for assets.

Use the string 'mondoo-operator-' plus the cluster's UID to uniquely
mark assets that have been discovered/scanned by the mondoo-operator.

* attempt garbage collection after successful scans

Introduce a 'garbage-collect' subcommand to ease testing/development,
and to allow the 'k8s-scan' subcommand to use this for garbage
collection purposes.

When a scan has run cleanly, make a GarbageCollectAssets() call for
resources that are older than 2 hours (for k8s assets), and older
than 2 days (for container-image assets).

For now, hide the functionality behind the
FEATURE_ENABLE_GARBAGE_COLLECTION feature flag.


joelddiaz:
do not merge until the necessary pieces of the Mondoo API have been pushed to prod

cmd/mondoo-operator/garbage_collect/cmd.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+		return err
+	}
+
+	return nil
+}
+
+func buildOlderThanTimestamp(olderThanString string) (string, error) {
+	var timestamp time.Time
+	if strings.HasSuffix(olderThanString, "m") {
```
czunker, line 120: The time package can parse durations: `time.ParseDuration`.

And it can also parse more complicated strings into durations, e.g., "2h10m42s". This makes it more flexible for us.

cmd/mondoo-operator/k8s_scan/cmd.go
```diff
@@ -74,6 +78,19 @@ func init() {
 			return err
 		}
 
+		// If scanning successful, now attempt some cleanup of older assets
+		if feature_flags.GetEnableGarbageCollection() && *setManagedBy != "" && *cleanupOlderThan != "" {
+			platformRuntime := garbage_collect.RUNTIME_KUBERNETES_CLUSTER
```
czunker, line 83: I'm not 100% sure, so: This works because image scan and k8s resource scan are seperate scans?

controllers/k8s_scan/container_image/deployment_handler.go
```diff
@@ -60,8 +60,14 @@ func (n *DeploymentHandler) syncCronJob(ctx context.Context) error {
 		return err
 	}
 
+	clusterUid, err := k8s.GetClusterUID(ctx, n.KubeClient, logger)
```
czunker, line 63: Nice. This way we can cleanup the images also when they are used in other clusters.

controllers/nodes/resources.go
```diff
@@ -205,7 +205,8 @@ func Inventory(node corev1.Node, integrationMRN, clusterUID string, m v1alpha2.M
 							PlatformId: fmt.Sprintf("//platformid.api.mondoo.app/runtime/k8s/uid/%s/node/%s", clusterUID, node.UID),
 						},
 					},
-					Labels: map[string]string{},
+					Labels:    map[string]string{},
+					ManagedBy: "mondoo-operator-" + clusterUID,
```
czunker, line 209: Is this just for completeness or do we also cleanup nodes?

controllers/k8s_scan/resources.go
```diff
@@ -37,12 +37,23 @@ func CronJob(image, integrationMrn string, m v1alpha2.MondooAuditConfig) *batchv
 
 		// The job runs hourly and we need to make sure that the previous one is killed before the new one is started so we don't stack them.
 		"--timeout", "55",
+		// Cleanup any resources more than 2 hours old
+		"--cleanup-assets-older-than", "2h",
```
czunker, line 41: Just wondering: Should we make the duration configurable?

controllers/k8s_scan/container_image/resources.go
```diff
@@ -40,12 +40,23 @@ func CronJob(image, integrationMrn string, m v1alpha2.MondooAuditConfig) *batchv
...
 		"--timeout", "1430",
+		// For a daily schedule, we will delete resources that are older than 2 days old
+		"--cleanup-assets-older-than", "48h",
 	}
 
 	if integrationMrn != "" {
 		containerArgs = append(containerArgs, []string{"--integration-mrn", integrationMrn}...)
 	}
 
+	// use the clusterUid to uniquely set the managedBy field on assets managed by this instance of the mondoo-operator
```
imilchev, line 51: why aren't we using the integrationMrn for this?

controllers/k8s_scan/resources.go
imilchev: I think at some point we would want that but it's probably good enough for first version

cmd/mondoo-operator/garbage_collect/cmd.go
joelddiaz: Nice!

cmd/mondoo-operator/k8s_scan/cmd.go
joelddiaz: The idea is that since only the daily CronJob sets the --scan-container-images flag, we can know when to cleanup container-images vs "the rest" of the assets.

controllers/k8s_scan/container_image/deployment_handler.go
joelddiaz: The plan was that with the container-image-scan-period * 2 as the "scan-older-than" filter, even if clusterA lost containerImageX, clusterB which still is scanning/updating containerImageX will be the "owner" of containerImageX before clusterA's garbage collection would ever try to garbage collect the previously shared container-image asset.

This approach does have the odd behavior where containerImageX will ping-pong on the ManagedBy field depending on which cluster scanned it most recently, but that is to be expected since the container-image assets are essentially shared across all Mondoo clients running in the same Mondoo Space.

controllers/nodes/resources.go
joelddiaz: As I was implementing this PR I noticed that we could set the ManagedBy field for the node resources as well, but this PR does nothing to attempt to cleanup the node resources. We are just letting the Mondoo client process "claim" the node assets with this.

Off the top of my head, if I were doing an implementation of garbage collection for nodes, I would probably create a second container to run alongside the node-scanning CronJob container, that would run the 'garbage-collect' container image/subcommand to specifically trigger the node-asset-type cleanup. But then as I write this that means that we would trigger node-cleanup multiple times (when the node's CronJob trigger to scan each individual node), so perhaps a dedicated CronJob just for node cleanup might make sense...

controllers/k8s_scan/container_image/resources.go
joelddiaz: That is what I initially implemented, but then I noticed since we technically allow setting up k8s scanning without an integration, it seemed unnecessary to require setting up a k8s-integration just to have asset garbage collection.

joelddiaz:
switched to using time.ParseDuration() and rebased/resolved-conflicts

czunker (approved):

PR #570, opened 2022-09-21
⭐️ Implement scanning AdmissionReview resources

imilchev:
- [ ] Implement RunAdmissionReview scan API endpoint
- [ ] Delete RunKubernetesManifest scan API endpoint
- [ ] Add feature flag for discovering admission review assets (currently, doesn't work because of a bug I introduced in cnquery which is fixed by https://github.com/mondoohq/cnquery/pull/77)


czunker (approved):
LGTM

PR #579, opened 2022-09-27
🧹 Use deployments instead of pods for integration tests

imilchev:
We added controls to flag individual pods that are not part of another deployment so our integration tests are now failing. We also need to fix:
- Pods should not run with NET_RAW capability
- Pods should not run with SYS_ADMIN capability

They shouldn't flag pods that don't explicitly disable these.

### DIFF
--- a/tests/integration/audit_config_base_suite.go
+++ b/tests/integration/audit_config_base_suite.go
@@ -579,13 +579,13 @@ func (s *AuditConfigBaseSuite) getFailingDeployment() *appsv1.Deployment {
 	labels := map[string]string{
 		"admission-result": "fail",
 	}
-	pod := s.getPassingDeployment().DeepCopy()
-	pod.ObjectMeta.Name = "failing-deployment"
-	pod.ObjectMeta.Labels = labels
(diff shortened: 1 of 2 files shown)


tests/integration/audit_config_base_suite.go
```diff
@@ -506,97 +506,110 @@ func (s *AuditConfigBaseSuite) disableContainerImageResolution() func() {
...
-	pod := s.getPassingPod().DeepCopy()
-	pod.ObjectMeta.Name = "failing-pod"
+	pod := s.getPassingDeployment().DeepCopy()
+	pod.ObjectMeta.Name = "failing-deployment"
 	pod.ObjectMeta.Labels = labels
-	pod.Spec.Containers[0].SecurityContext = &corev1.SecurityContext{
+	pod.Spec.Template.Spec.Containers[0].SecurityContext = &corev1.SecurityContext{
 		Privileged: pointer.Bool(true),
 	}
 	return pod
```
joelddiaz, line 588: s/pod/deployment/ for the variable name

joelddiaz (approved):

PR #580, opened 2022-09-27
turn on garbage collection by default

joelddiaz:
If you want to disable it unset
FEATURE_ENABLE_GARBAGE_COLLECTION in the operator Deployment.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -35,6 +35,8 @@ spec:
         env:
         - name: FEATURE_DISCOVER_WORKLOADS
           value: "1"
+        - name: FEATURE_ENABLE_GARBAGE_COLLECTION
+          value: "1"
         image: controller:latest
         imagePullPolicy: IfNotPresent
         name: manager


vjeffrey (dismissed):

pkg/feature_flags/feature_flags.go
```diff
@@ -63,9 +67,19 @@ func setGlobalFlags(k, v string) {
...
+			enableGarbageCollection = false
+		} else {
+			enableGarbageCollection = true
+		}
 	case "FEATURE_ENABLE_ADMISSION_REVIEW_DISCOVERY":
 		enableAdmissionReviewDiscovery = true
 	}
 }
+
+func setDefaults() {
```
vjeffrey, line 81: 👍

vjeffrey (dismissed):

pkg/feature_flags/feature_flags.go
```diff
@@ -63,9 +67,19 @@ func setGlobalFlags(k, v string) {
 		enablePodDiscovery = true
 	case "FEATURE_DISCOVER_WORKLOADS":
 		enableWorkloadDiscovery = true
-	case "FEATURE_ENABLE_GARBAGE_COLLECTION":
-		enableGarbageCollection = true
+	case FeatureGarbageCollection:
```
imilchev, line 70: I think this change is not really needed. The way I enabled workload discovery is by just setting the env var to "1" in the deployment manifest. https://github.com/mondoohq/mondoo-operator/blob/main/config/manager/manager.yaml#L36-L37 I think that's much cleaner than altering the code

joelddiaz: I suppose that is a lot simpler than my approach. I'll update/rework the PR.

imilchev (approved):
Sweet! Then when we know it works all good, we just drop the flags completely!

PR #584, opened 2022-10-04
🧹 Stabilize integration tests

czunker:
This should prevent  errors we saw lately in our integration tests.


czunker:
This errors happens more often the last weeks:
```
2022-09-27T14:40:04.774Z	INFO	integration/audit_config_base_suite.go:592	Create a Deployment which should pass.
    audit_config_base_suite.go:593: 
        	Error Trace:	/home/runner/work/mondoo-operator/mondoo-operator/tests/integration/audit_config_base_suite.go:593
        	            				/home/runner/work/mondoo-operator/mondoo-operator/tests/integration/audit_config_base_suite.go:267
        	            				/home/runner/work/mondoo-operator/mondoo-operator/tests/integration/audit_config_base_suite.go:213
        	            				/home/runner/work/mondoo-operator/mondoo-operator/tests/integration/audit_config_test.go:41
        	Error:      	Received unexpected error:
        	            	Internal error occurred: failed calling webhook "policy.k8s.mondoo.com": failed to call webhook: Post "[https://mondoo-client-webhook-service.mondoo-operator.svc:443/validate-k8s-mondoo-com?timeout=10s](https://mondoo-client-webhook-service.mondoo-operator.svc/validate-k8s-mondoo-com?timeout=10s)": EOF
        	Test:       	TestAuditConfigSuite/TestReconcile_AdmissionEnforcing
        	Messages:   	Failed to create Deployment which should pass.
2022-09-27T14:40:04.780Z	INFO	integration/audit_config_base_suite.go:597	Create a Deployment which should be denied in enforcing mode.
```

the easy fix would be to increase the sleep period. This hopefully makes it more stable without spending too much time sleeping when everything is fine.

imilchev (commented):
Does that work in all cloud distros of k8s? I know in some of them the nodes are protected with a firewall so node ports might not work out of the box

czunker:
> I started a run of the cloud tests. Let's see what happens https://github.com/mondoohq/mondoo-operator/actions/runs/3183884050

Thanks. They all failed. The reasons seems strange. I added some debug output to the branch and started a new run:
https://github.com/mondoohq/mondoo-operator/actions/runs/3184317301

At least for EKS, I added a firewall rule. AKS and GKE didn't seem to have any configured in terraform. But as you said, let's see what happens.

czunker:
EKS is working now.

AKS and GKE aren't. Perhaps the firewall issue? Will check how to add rules for the nodes.

czunker:
GKE needs additional permissions.

imilchev:
I feel like the node port service might not be the best solution for solving this issue. We are changing too much from the cloud config to enable this. I know there were some issues with changing the firewall policy of AKS which could take 30mins or so to happen. Why not avoid doing this and just use `kubectl port-forward` to forward the service to a local port and use that instead?

czunker:
Ok, now with `kubectl port-forward`. The `portforward` package of the client had problems with minikube. So, let's try this.

This looks like a hack. I'm open to any suggestions...

But before I improve this, I'd like to see it working with minikube, k3s and our cloud tests.

czunker:
Ok, this is now working locally, in GitHub with minikube, k3s, and clouds:
https://github.com/mondoohq/mondoo-operator/actions/runs/3190312931

imilchev (approved):
sweet!

imilchev (approved):

PR #585, opened 2022-10-04
🐛 Fix `MondooAuditConfig` Status Reconcile

### LINKED ISSUE #433: Re-write mondooauditconfig Reconcile in a re-entrant way
**Describe the bug**
The current implementation might create race-conditions when not run single-threaded.

See comment here: https://github.com/mondoohq/mondoo-operator/pull/395#discussion_r913832053

**To Reproduce**
-

**Expected behavior**
Reconcile should only do one thing and then re-iterate.

**Screenshots**
-

**Additional context**
Perhaps also defer the status update, as done here:
https://github.com/mondoohq/mondoo-operator/pull/408/files#diff-95f7e0c1611028f1d111b1f9ebada14e1b8919e068ccb23fbd3cd37b19d7b797R181

czunker:
This should prevent any race-conditions, which were previously possible.

Fixes #433

### DIFF
--- a/controllers/mondooauditconfig_controller_test.go
+++ b/controllers/mondooauditconfig_controller_test.go
@@ -381,10 +381,10 @@ func TestMondooAuditConfigStatus(t *testing.T) {
 			Namespace: testNamespace,
 		},
 	})
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
 	err = fakeClient.Get(context.TODO(), client.ObjectKeyFromObject(mondooAuditConfig), mondooAuditConfig)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
 	assert.NotEmptyf(t, mondooAuditConfig.Status, "Status shouldn't be empty")
 	assert.NotEmptyf(t, mondooAuditConfig.Status.ReconciledByOperatorVersion, "ReconciledByOperatorVersion shouldn't be empty")
--- a/controllers/mondooauditconfig_controller_test.go
+++ b/controllers/mondooauditconfig_controller_test.go
@@ -372,7 +372,6 @@ func TestMondooAuditConfigStatus(t *testing.T) {
 	reconciler := &MondooAuditConfigReconciler{
 		MondooClientBuilder: testMondooClientBuilder,
 		Client:              fakeClient,
-		MondooAuditConfig:   mondooAuditConfig,
 		ScanApiStore:        scanApiStore,
 	}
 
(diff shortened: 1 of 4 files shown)


imilchev:
What does this fix exactly?

czunker:
> What does this fix exactly?

@joelddiaz suspected some race conditions here: https://github.com/mondoohq/mondoo-operator/pull/395#discussion_r913832053

Part of the fix was already done with https://github.com/mondoohq/mondoo-operator/pull/408. But that introduced an update to 
`ReconciledByOperatorVersion`, which could indicate success when not everything was reconciled.

So, basically, with #395, I updated and saved the `MondooAuditConfig` not in an operator-like manner.

controllers/mondooauditconfig_controller_test.go
```diff
@@ -384,9 +383,12 @@ func TestMondooAuditConfigStatus(t *testing.T) {
 	})
 	assert.NoError(t, err)
 
-	assert.NotEmptyf(t, reconciler.MondooAuditConfig.Status, "Status shouldn't be empty")
-	assert.NotEmptyf(t, reconciler.MondooAuditConfig.Status.ReconciledByOperatorVersion, "ReconciledByOperatorVersion shouldn't be empty")
-	assert.Equalf(t, reconciler.MondooAuditConfig.Status.ReconciledByOperatorVersion, version.Version, "expected versions to be equal")
+	err = fakeClient.Get(context.TODO(), client.ObjectKeyFromObject(mondooAuditConfig), mondooAuditConfig)
+	assert.NoError(t, err)
```
joelddiaz, line 387: s/assert/require/ since the asserts() that follow reach into the mondooAuditConfig which is the same mondooAuditConfig as in line 365 (so I could see us getting some weird assertion errors that are about what testMondooAuditConfig() returned instead of the results of the Get()

imilchev:
> > What does this fix exactly?
> 
> @joelddiaz suspected some race conditions here: [#395 (comment)](https://github.com/mondoohq/mondoo-operator/pull/395#discussion_r913832053)
> 
> Part of the fix was already done with #408. But that introduced an update to `ReconciledByOperatorVersion`, which could indicate success when not everything was reconciled.
> 
> So, basically, with #395, I updated and saved the `MondooAuditConfig` not in an operator-like manner.

but it seems like this is only a problem if we have multi-threaded operator, which we don't AFAIK. That's why I was wondering why this is a problem

joelddiaz:
> but it seems like this is only a problem if we have multi-threaded operator, which we don't AFAIK. That's why I was wondering why this is a problem

Yeah, I don't see anywhere where we change the MaxConcurrentReconciles on the controller, so we're effectively single-threaded even in the case of two MondooAuditConfig resources.

But I would consider it concerning to store temporary/ephemeral things into the shared instantiation of `MondooAuditConfigReconciler` (assuming it really is a single instance of it that the manager sets up).

controllers/mondooauditconfig_controller_test.go
czunker: Thanks. Hopefully, I'll remember this from now on. It's not the first time you find this in my changes.

PR #588, opened 2022-10-05
🐛 Fix token for checkout

czunker:
This fixes the error message:
'Unexpected input(s) 'github-token', valid inputs are ['repository', 'ref', 'token', 'ssh-key', 'ssh-known-hosts', 'ssh-strict', 'persist-credentials', 'path', 'clean', 'fetch-depth', 'lfs', 'submodules', 'set-safe-directory']

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -35,7 +35,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
           persist-credentials: false
           fetch-depth: 0 # fetch is nneded for "git tag --list" in the Makefile
       - name: Import environment variables from file
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -18,7 +18,9 @@ env:
 
 # https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
 permissions:
-  contents: read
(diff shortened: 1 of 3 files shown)


czunker:
Ok, more things to fix. By default `workflow_run` will checkout the default branch: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

The `workflow_run` jobs will not appear in the PR overview: https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-796636235

joelddiaz:
> Ok, more things to fix. By default `workflow_run` will checkout the default branch: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run
> 
> The `workflow_run` jobs will not appear in the PR overview: [dependabot/dependabot-core#3253 (comment)](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-796636235)
> 
So when looking whether a PR has successfully passed or not we would need to dive into the actions results for the repo?

joelddiaz (approved):

czunker:
> > Ok, more things to fix. By default `workflow_run` will checkout the default branch: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run
> > The `workflow_run` jobs will not appear in the PR overview: [dependabot/dependabot-core#3253 (comment)](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-796636235)
> 
> So when looking whether a PR has successfully passed or not we would need to dive into the actions results for the repo?

Yes

PR #593, opened 2022-10-07
🧹 Re-work PR workflows

czunker:
This will bring back the integration and security tests into the PR.

It will also allow us dependabot and fork PRs. But we need to apply a label, before the tests start.

The label is also directly removed after the workflow starts, to prevent later commits from just starting the tests again.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -23,7 +23,7 @@ jobs:
       pull-requests: write
     steps:
       - name: remove labels
-        uses: andymckay/labeler@v1
+        uses: andymckay/labeler@1.0.4
         with:
           remove-labels: "ok to test"
   unit-tests:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -19,7 +19,7 @@ jobs:
     name: Check label
     runs-on: ubuntu-latest
     if: |
-      !github.event.pull_request.head.repo.fork
+      (!github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]')
       || (github.actor == 'dependabot[bot]' && contains(github.event.pull_request.labels.*.name, 'ok to test'))
       || (github.event.pull_request.head.repo.fork && contains(github.event.pull_request.labels.*.name, 'ok to test'))
     permissions:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,5 +1,6 @@
 name: Run Test
 on:
+  pull_request:
   pull_request_target:
     types: [opened, synchronize, reopened, labeled]
   push:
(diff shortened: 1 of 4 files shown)


czunker:
Security test should pass when this is merged: https://github.com/mondoohq/mondoo-operator/pull/592

.github/workflows/tests.yaml
```diff
@@ -9,32 +10,39 @@ on:
...
-
-      - run: make test/ci
-      - uses: actions/upload-artifact@v3  # upload test results
-        if: success() || failure()        # run this step even if previous step failed
-        with:
-          name: test-results
-          path: unit-tests.xml
-   
\ No newline at end of file
+          remove-labels: "ok to test"
```
joelddiaz, line 32: this does mean if we run into a flaky test, we just have to re-label to get a second run attempt?

.github/workflows/tests.yaml
```diff
@@ -1,40 +1,51 @@
...
 jobs:
-  unit-tests:
+  check-label:
+    name: Check label
     runs-on: ubuntu-latest
-    name: Unit tests
+    if: |
+      !github.event.pull_request.head.repo.fork
+      || (github.actor == 'dependabot[bot]' && contains(github.event.pull_request.labels.*.name, 'ok to test'))
+      || (github.event.pull_request.head.repo.fork && contains(github.event.pull_request.labels.*.name, 'ok to test'))
```
joelddiaz, line 24: help me understand this a bit. the way i read these conditions is
if not_a_fork || dependabot AND is_labeled || is_a_fork and is_labeled

aren't dependabot PRs already not a fork? what does the middle check accomplish?

.github/workflows/tests.yaml
czunker: Would also work. The idea was more like:
- fork PR is created
- we check it, seems good
- we add a label so the tests run
- a commit is added to grap secrets <= this would run, when the label is still present

.github/workflows/tests.yaml
joelddiaz: right, dependabot isn't a fork, so doesn't it just go down the path of the 'not_a_fork' check?

if we're using is_a_fork as a test on a PR, then how can we tell the difference between our PRs and dependabot PRs?

.github/workflows/tests.yaml
joelddiaz: Yes, anytime the workflow runs it clears the label. So to have a chance at re-running a workflow, we essentially need to re-label it. To be clear this is pretty much how I expected it to work, I'm just double checking that my reading of the workflow file is correct.

.github/workflows/tests.yaml
czunker: Ok, now I think I understand your question. I added an additional check to make sure non-fork PRs aren't created by dependabot for the first case.

.github/workflows/tests.yaml
czunker: Correct.

PR #602, opened 2022-10-12
🐛 Use separate entrypoints to workflows

czunker:
This prevents duplicate workflow runs for non-forks/dependabot PRs

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -13,22 +13,9 @@ permissions:
   contents: read
 
 jobs:
-  check-label:
-    name: Check label
-    runs-on: ubuntu-latest
-    if: |
-      !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]'      
-    permissions:
-      issues: write
-      pull-requests: write
-    steps:
-      - name: remove labels
-        uses: andymckay/labeler@1.0.4
-        with:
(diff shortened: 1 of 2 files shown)


czunker:
This happened before:

image

.github/workflows/tests.yaml
```diff
@@ -19,9 +17,7 @@ jobs:
     name: Check label
     runs-on: ubuntu-latest
     if: |
-      (!github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]')
-      || (github.actor == 'dependabot[bot]' && contains(github.event.pull_request.labels.*.name, 'ok to test'))
-      || (github.event.pull_request.head.repo.fork && contains(github.event.pull_request.labels.*.name, 'ok to test'))
+      !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]'
```
joelddiaz, line 20: with these two separate paths for testing (1 for dependabot / forked PRs and 1 for people who can commit directly to the repo), do we even need the label checking for this codepath for people who created a non-forked PR and are not the 'dependabot[]bot]' user?

czunker: Good catch. Thanks. I removed the steps and moved the "if" to the units tests directly.

joelddiaz (approved):

PR #603, opened 2022-10-12
🐛 Drop all capabilities

### LINKED ISSUE #599: Mondoo operator should run w/o NET_RAW capability
**Describe the bug**
Mondoo operator should run w/o NET_RAW capability

**To Reproduce**
1. Install the operator
2. Look at the Kubernetes Cluster and Workload Security policy for the operator itself

**Expected behavior**
Passes

**Screenshots**
image

czunker:
Fixes #599

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -43,6 +43,9 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
+          capabilities:
+            drop:
+            - ALL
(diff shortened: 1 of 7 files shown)


czunker:
To fix #599 we could have only dropped NET_RAW, but we do not need the other caps, so why not drop them also.

This is a list of default capabilities: https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities

joelddiaz (approved):

PR #606, opened 2022-10-14
Update cla.yaml

benr:


.github/workflows/cla.yaml
```diff
@@ -25,4 +25,4 @@ jobs:
           remote-organization-name: mondoohq
           # branch should not be protected
           branch: 'main'
-          allowlist: mondoo-tools
+          allowlist: mondoo-tools, *bot
```
joelddiaz, line 28: Does this mean that any and all users ending in `bot` are on the allowlist? Would putting in the full username of `depandabot` (or is it `dependabot[bot]` as that seems like the full "github actor" name) restrict this a bit more?

It even seems like regular github users cannot have brackets in their names, so perhaps `*[bot]` would cover things a little more broadly.

benr: This change is a half step to see whats coming back, it's unclear if the full name is dependabot or dependabot[bot].  Another change is coming this afternoon to update the message, so it'll be updated to it's full name to avoid any conflict with "evilbotuser32" in the future. ;)

joelddiaz (approved):

benr:
Follow up PR is https://github.com/mondoohq/mondoo-operator/pull/607

PR #612, opened 2022-10-18
🧹 Use edge config for edge tests

czunker:


czunker:
We need the new secret `MONDOO_CLIENT_EDGE` before tests can pass.

joelddiaz (commented):
If I'm reading this correctly, we can run our usual defaults (v6-based client) against prod, and the edge builds now will run against edge. :100:

joelddiaz (approved):

PR #613, opened 2022-10-18
⭐️ Use Mondoo client v7 instead of v6

imilchev:
We can only merge this one after v7 has been released.


czunker:
Superseded by https://github.com/mondoohq/mondoo-operator/pull/625

PR #616, opened 2022-10-18
🧹 webhook shouldn't use default serviceaccount

joelddiaz:
it is considered bad form to use the 'default' ServiceAccount in any Namespace. give the webhook its own ServiceAccount (with no special permissions associated with it), and use that webhook.

Extend the MondooAuditConfig for the case where a MondooAuditConfig is installed somewhere besides 'mondoo-operator' so that a user can provide/specify their own SerivceAccount in whatever Namespace the MondooAuditConfig was installed onto.

### DIFF
--- a/controllers/admission/resources.go
+++ b/controllers/admission/resources.go
@@ -160,6 +160,7 @@ func WebhookDeployment(ns, image string, m mondoov1alpha2.MondooAuditConfig, int
 					SecurityContext: &corev1.PodSecurityContext{
 						RunAsNonRoot: pointer.Bool(true),
 					},
+					ServiceAccountName:            m.Spec.Admission.ServiceAccountName,
 					TerminationGracePeriodSeconds: pointer.Int64(10),
 					Volumes: []corev1.Volume{
 						{
--- a/controllers/admission/resources.go
+++ b/controllers/admission/resources.go
@@ -160,6 +160,8 @@ func WebhookDeployment(ns, image string, m mondoov1alpha2.MondooAuditConfig, int
 					SecurityContext: &corev1.PodSecurityContext{
 						RunAsNonRoot: pointer.Bool(true),
 					},
+					// service account used during the initial bringup of the webhook
+					// by the supporting controller-runtime packages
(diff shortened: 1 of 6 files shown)


controllers/admission/resources.go
```diff
@@ -39,6 +39,10 @@ const (
 	// manualTLSAnnotationKey is for when there is no explicit 'injection-style' defined in
 	// the MondooAuditConfig. We treat this to mean that the user will provide their own certs.
 	manualTLSAnnotationKey = "mondoo.com/tls-mode"
+
+	// webhookServiceAccountName is the name of the generated ServiceAccount for use by
+	// the webhook
+	webhookServiceAccountName = "mondoo-operator-webhook"
```
imilchev, line 45: why is a new account needed? Can't we just not mount the service account token of the default account?

joelddiaz: i tried that first. look at the name of my branch :)

but it turns out that on startup of the webhook does go through a codepath that wants access to a kubeconfig
```
2022-10-18T23:13:21Z    INFO    webhook setting up manager
2022-10-18T23:13:21Z    ERROR   controller-runtime.client.config        unable to get kubeconfig        {"error": "invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable", "errorCauses": [{"error": "no configuration has been provided, try setting KUBERNETES_MASTER environment variable"}]}
sigs.k8s.io/controller-runtime/pkg/client/config.GetConfigOrDie
        /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/client/config/config.go:153
```

imilchev: Yeah, I saw it and that made me even more curious 😁 perhaps add a comment so that it is clear why that account is actually used. I suppose the webhook creates a kube client, when it starts but we don't really use it right now.

joelddiaz: Looks like it is right around here https://github.com/mondoohq/mondoo-operator/blob/main/cmd/mondoo-operator/webhook/cmd.go#L56

czunker (approved):
LGTM

PR #623, opened 2022-10-20
annotate the mondoo-operator namespace

joelddiaz:
to deal with PSS checks


czunker (approved):
:+1:

PR #633, opened 2022-10-21
⭐️ include/exclude Namespaces filtering

joelddiaz:
Extend MondooAuditConfig to allow specifying Namespaces to include or
exclude. Specifying anything in Include will override anything listed in
Exclude (as Include effectively excludes everything that isn't in the
Include list).

ensure the following components are aware of include/exclude namespaces:

k8s scanning binary receives --namespaces and --namespaces-exclude param
and passes them down via inventory options when performing the periodic
scan Jobs.

the webhook can be launched with --namespaces --namespaces-exclude to
also affect whether a resource the webhook has been notified about
should be skipped or scanned.

the scan_api_store stores the include/exclude data to be used for
deciding whether or not to scan the batches of items is processes.

some of the decisions on whether to scan or not could have been pushed
down into the mondoo scan-api Pod itself, but it was easy enough to just
filter resources out without needing to trigger any kind of scanning by
the scan-api Service.


joelddiaz:
Note that for the periodic full cluster scan Jobs, it requires a Mondoo client that includes the functionality that recently arrived in https://github.com/mondoohq/cnquery/pull/357 . Without that, the previous clients still perform a full cluster scan and just ignore the namespaces / namespaces-exclude options.

config/samples/k8s_v1alpha2_mondooauditconfig.yaml
```diff
@@ -30,4 +30,11 @@ spec:
       tag: latest
     # could be "permissive" or "enforcing"
     mode: permissive
-  
+  filtering:
+    namespaces:
+      exclude:
```
czunker, line 35: Should we include this in the example? From what I understood, this `exclude` will not be applied but overridden by the `include`. Correct?

controllers/resource_monitor/debouncer/debouncer.go
```diff
@@ -66,9 +69,18 @@ func (d *debouncer) Start(ctx context.Context) {
 
 			for res := range d.resources {
 				for _, c := range clients {
-					logger.Info("Reconciling change", "request", res, "integration-mrn", c.IntegrationMrn)
-					if _, err := c.Client.ScheduleKubernetesResourceScan(ctx, c.IntegrationMrn, res); err != nil {
-						logger.Error(err, "Failed to schedule resource scan", "request", res)
+					fields := strings.Split(res, ":")
+					if len(fields) != 3 {
```
czunker, line 73: I think this changed lately. Non-namespaced objects now only look like `kind:name`.

czunker: https://github.com/mondoohq/cnquery/pull/318

pkg/utils/utils.go
```diff
@@ -0,0 +1,27 @@
...
+Copyright 2022 Mondoo, Inc.
+
+This Source Code Form is subject to the terms of the Mozilla Public
+License, v. 2.0. If a copy of the MPL was not distributed with this
+file, You can obtain one at https://mozilla.org/MPL/2.0/.
+*/
+
+package utils
+
+func AllowNamespace(namespace string, includeNamespaces, excludeNamespaces []string) bool {
```
czunker, line 11: I'm not 100% sure, but something feels strange here. Not explicitly at this line, but for the namespace parsing.

What I understood is, you check the namespaces in the operator to not scan resources that are not on the list. But you also parse the namespace filter list to the client. Is the second step needed?

joelddiaz: For the "dynamic" scanning like the webhook and the resource_monitor, we know the namespace on the resource right on the spot, so there isn't really any need to push work to the scanAPI as we can just drop the processing of the resource right on the spot.

For the scheduled scans that scan the entire cluster, we then pass the namespace info to the scan API so that there is client-side filtering of the resources.

controllers/resource_monitor/debouncer/debouncer.go
joelddiaz: Since the resource_monitor doesn't watch Namespace resources, we should always be getting notified about resources with a full type:namespace:resourceName set. (At least for how the resource_monitor is written today...)

config/samples/k8s_v1alpha2_mondooauditconfig.yaml
joelddiaz: Yes. I included it in the sample as just a quick reference for the kinds of fields you can fill out when defining a MondooAuditConfig. I can remove either the include or exclude list, or I can put a comment on the YAML explaining that setting the include list overrides the contents of the exclude list. WDYT?

joelddiaz:
rebased after #637 merged

tas50:
LGTM to me from a naming perspective

pkg/utils/utils.go
czunker: Ah. Thanks.

config/samples/k8s_v1alpha2_mondooauditconfig.yaml
czunker: I think a comment would be nice. I assume most will copy&paste, enter their namespaces, and later wonder what happened. This way they get a hint.

czunker (dismissed):

czunker (approved):

Issue #621, opened 2022-10-20
Wrong artifact tagged as latest


czunker:
**Describe the bug**
During a release the Helm chart got tagged as latest.

Instead the actual operator should have been tagged as latest.

This happened for the v1.6.5 release. Perhaps also before and we just didn't notice.

This error results in errors for the integrations install via Web UI.

**To Reproduce**
???

**Expected behavior**
The Helm chart shouldn't be tagged as latest.

**Screenshots**

**Additional context**
> ... operations were impacted for the last 5 hours because https://github.com/mondoohq/mondoo-operator/releases/tag/v1.6.5 was not marked as "latest"... instead https://github.com/mondoohq/mondoo-operator/releases/tag/mondoo-operator-1.6.5 was marked latest.

czunker:
Sounds like we have to make sure the helm chart is released before the manifest:
https://github.blog/changelog/2022-10-21-explicitly-set-the-latest-release/
Then, the manifest should get the latest tag.

PR #634, opened 2022-10-24
🐛 Ensure manifest gets latest tag

czunker:

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -362,6 +362,8 @@ jobs:
     uses: ./.github/workflows/release-manifests.yaml
     needs:
       - push-virtual-tag
+# this should ensure the manifest is tagged latest, which is required for the install automation
+      - release-helm
 
   # publish helm chart after the release of container images is complete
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -362,6 +362,8 @@ jobs:
     uses: ./.github/workflows/release-manifests.yaml
     needs:
       - push-virtual-tag
+# this should ensure the manifest is tagged latest, which is required for the install automation
+      - release-helm
```
imilchev, line 366: could you elaborate a bit more what the problem is here?

czunker: Oh, the PR is missing the link to the issue: https://github.com/mondoohq/mondoo-operator/issues/621

Basically, the helm chart got the latest tag on some releases lately. But install.mondoo.com relies on the latest tag being applied to the release containing the manifest. As both releases are built from the same commit, the most current release will get the latest tag.

joelddiaz (approved):
nice. it was getting old having to go manually re-label things as latest...

PR #648, opened 2022-10-26
🧹 Retrieve real operator version before cloud tests start

imilchev:
The problem was that we do not map 1:1 each container tag to an operator version. For example, if we pull the container image with a tag `main` the operator version inside is not `main`. With this change before the test suite is started we retrieve the operator version from inside the container. In this way it should be always the correct one


czunker (approved):
Nice!

PR #649, opened 2022-10-26
🐛 Fix debug output

czunker:


czunker:
The debug-event test will not succeed here because of `pull_request_target`.

imilchev (approved):

PR #650, opened 2022-10-27
🐛 Check for user login

czunker:
When we apply the label, the actor is changed to our user. Same for the previous login field. That prevented the checks from starting. Only a `@dependabot rebase` or `@dependabot recreate` triggered the checks again.

This field holds the dependabot user, even when we apply a label.


czunker:
It's hard to read, but the user names are in there: https://github.com/mondoohq/mondoo-operator/actions/runs/3335515666/jobs/5519571417

imilchev (approved):

PR #651, opened 2022-10-27
🧹 Set workloads we want to discovery with the webhook

czunker:


czunker:
This is a follow-up to this discussion: https://github.com/mondoohq/cnquery/pull/305#discussion_r997066932

It is related to https://github.com/mondoohq/cnquery/pull/389

pkg/webhooks/handler/webhook.go
```diff
@@ -162,10 +162,8 @@ func (a *webhookValidator) Handle(ctx context.Context, req admission.Request) (r
 	}
 
 	scanJob.Discovery = &inventory.Discovery{}
-	if feature_flags.GetEnableWorkloadDiscovery() {
```
imilchev, line 165: If you do this, then remove the feature flag as well. Also remove the env var that is set for the operator in the operator manifest. Then we have completely cleaned up the workload discovery feature flag

czunker: I removed the workload and pod discovery feature flags.

pkg/mondooclient/client.go
```diff
@@ -267,9 +266,20 @@ func (s *mondooClient) ScanKubernetesResources(ctx context.Context, scanOpts *Sc
 						Connections: []inventory.TransportConfig{
 							{
 								Backend: inventory.TransportBackend_CONNECTION_K8S,
+								Options: map[string]string{
+									"all-namespaces": "true",
```
czunker, line 270: @imilchev This would override the changes you just made here: https://github.com/mondoohq/mondoo-operator/pull/657/files#diff-b212dccf6feabc654e6f33451f64d9470e12bedc37180b157d5b5d710b202c75L291

czunker:
Strang `lint` is flagging issues which didn't change and I do not get the errors when running locally with the same version.

czunker:
> Strange `lint` is flagging issues which didn't change and I do not get the errors when running locally with the same version.

After rebasing, I also got the issues locally.

pkg/mondooclient/client.go
```diff
@@ -337,6 +333,18 @@ func (s *mondooClient) ScheduleKubernetesResourceScan(ctx context.Context, integ
 								Options: map[string]string{
 									"k8s-resources": resourceKey,
 								},
+								Discover: inventory.Discovery{
+									Targets: []string{
```
imilchev, line 337: do we want to use this or use `auto` which is effectively the same thing?

czunker: Sounds like a good idea to me. I changed it here and L274

PR #657, opened 2022-10-28
🐛 Fix namespace filtering

imilchev:
We received an issue that namespace filtering does not work as expected. After debugging it I discovered it actually never worked because the wrong options were set in the inventory. This PR fixes the issue

### DIFF
--- a/pkg/mondooclient/client.go
+++ b/pkg/mondooclient/client.go
@@ -271,13 +271,13 @@ func (s *mondooClient) ScanKubernetesResources(ctx context.Context, scanOpts *Sc
 									// always disocver the cluster, this is needed for the asset explorer
 									Targets: []string{"clusters"},
 								},
+								Options: map[string]string{
+									"namespaces":         strings.Join(scanOpts.IncludeNamespaces, ","),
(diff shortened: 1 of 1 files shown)


preslavgerchev (approved):

pkg/mondooclient/client.go
```diff
@@ -286,11 +286,6 @@ func (s *mondooClient) ScanKubernetesResources(ctx context.Context, scanOpts *Sc
 
 	setIntegrationMrn(scanOpts.IntegrationMrn, &scanJob)
 
-	if scanOpts.ScanContainerImages || feature_flags.GetEnablePodDiscovery() || feature_flags.GetEnableWorkloadDiscovery() {
-		scanJob.Inventory.Spec.Assets[0].Connections[0].Options = make(map[string]string)
-		scanJob.Inventory.Spec.Assets[0].Connections[0].Options["all-namespaces"] = "true"
```
czunker, line 291: Don't we still need this? What happens when include and exclude are empty?

imilchev: As discussed we now do all-namespaces by default so this is not needed

PR #660, opened 2022-10-31
🐛 Fix restart of port-forward

### LINKED ISSUE #658: Restart of port-forward during integration tests fails
**Describe the bug**
```
2022-10-31T06:48:43.799Z	DEBUG	integration/audit_config_base_suite.go:810	Webhook is not available yet: Post "https://127.0.0.1:18443/validate-k8s-mondoo-com": dial tcp 127.0.0.1:18443: connect: connection refused
2022-10-31T06:48:43.799Z	DEBUG	integration/audit_config_base_suite.go:821	Trying to restart port-forward
2022-10-31T06:48:49.801Z	DEBUG	integration/audit_config_base_suite.go:810	Webhook is not available yet: Post "https://127.0.0.1:18443/validate-k8s-mondoo-com": dial tcp 127.0.0.1:18443: connect: connection refused
2022-10-31T06:48:49.801Z	DEBUG	integration/audit_config_base_suite.go:821	Trying to restart port-forward
2022-10-31T06:48:54.804Z	INFO	integration/audit_config_base_suite.go:788	Killing port-forward with pid: 20672
    audit_config_base_suite.go:277: 
        	Error Trace:	/home/runner/work/mondoo-operator/mondoo-operator/tests/integration/audit_config_base_suite.go:277
        	            				/home/runner/work/mondoo-operator/mondoo-operator/tests/integration/audit_config_base_suite.go:224
        	            				/home/runner/work/mondoo-operator/mondoo-operator/tests/integration/audit_config_test.go:41
        	Error:      	Received unexpected error:
make: *** [Makefile:147: test/integration/ci] Error 1
        	            	webhook not available: couldn't start kubectl port-forward: exec: already started
        	Test:       	TestAuditConfigSuite/TestReconcile_AdmissionEnforcing
        	Messages:   	Couldn't access Webhook via port-forward
2022-10-31T06:48:54.805Z	INFO	integration/audit_config_base_suite.go:278	Webhook should be working by now.
```

https://github.com/mondoohq/mondoo-operator/actions/runs/3359096924/attempts/1

**To Reproduce**

**Expected behavior**
Tests should re-create port-forward and pass.

czunker:
`cmd` cannot be reused:
https://pkg.go.dev/os/exec#Cmd

Fixes #658

### DIFF
--- a/tests/integration/audit_config_base_suite.go
+++ b/tests/integration/audit_config_base_suite.go
@@ -774,7 +774,7 @@ func (s *AuditConfigBaseSuite) checkWebhookAvailability() error {
 		return fmt.Errorf("couldn't start port-forward: %w", err)
 	}
 	// kubectl port-forwarding does not return but will run until interrupted
-	// We have to get ride of the port-forward at the end, because we need to create a new one for each test
+	// We have to get rid of the port-forward at the end, because we need to create a new one for each test
 	defer func() {
 		err := s.stopCmd(cmd)
 		if err != nil {
--- a/tests/integration/audit_config_base_suite.go
+++ b/tests/integration/audit_config_base_suite.go
@@ -8,7 +8,6 @@ import (
 	"net/http"
 	"os/exec"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/stretchr/testify/suite"
@@ -769,15 +768,7 @@ func (s *AuditConfigBaseSuite) checkWebhookAvailability() error {
 	// there is this package https://pkg.go.dev/k8s.io/client-go/tools/portforward
 	// But it seems this is a bit complicated in combination with minikube
 	// because of that we use kubectl directly
-	kubectlArgs := []string{
-		"-n",
-		webhookService.Namespace,
-		"port-forward",
-		"svc/" + webhookService.Name,
-		fmt.Sprintf("%d:%d", webhookLocalPort, webhookService.Spec.Ports[0].Port),
-	}
-
-	cmd := exec.Command("kubectl", kubectlArgs...)
+	cmd := s.createPortForwardCmd(webhookService)
 	err := cmd.Start()
 	if err != nil {
 		return fmt.Errorf("couldn't start port-forward: %w", err)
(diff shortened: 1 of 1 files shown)


tests/integration/audit_config_base_suite.go
```diff
@@ -799,43 +789,53 @@ func (s *AuditConfigBaseSuite) checkWebhookAvailability() error {
...
-		if err != nil {
-			zap.S().Debug("Port-forward is not running anymore: %v", err)
-			break
-		}
-		time.Sleep(1 * time.Second)
-	}
-	err = cmd.Start()
-	if err != nil {
-		return fmt.Errorf("couldn't start kubectl port-forward: %w", err)
+	_, err = cmd.Process.Wait()
```
joelddiaz, line 827: do we need to worry about a potential block here? or does Wait() have its own internal timeout?

tests/integration/audit_config_base_suite.go
```diff
@@ -799,43 +789,53 @@ func (s *AuditConfigBaseSuite) checkWebhookAvailability() error {
...
+			zap.S().Debug("Webhook is not available yet: ", webhookErr)
 			// this error will not recover itself over time, we have to re-connect
-			if strings.HasSuffix(err.Error(), "connection refused") {
-				err = s.restartCmd(cmd)
+			if strings.HasSuffix(webhookErr.Error(), "connection refused") {
+				zap.S().Debug("Trying to restart port-forward")
+				err := s.stopCmd(cmd)
+				if err != nil {
+					zap.S().Errorf("couldn't stop port-forward: %v\n", err)
+					continue
```
joelddiaz, line 808: does the continue here mean that the `kubectl port-forward` command is still hanging around? and if so, does a further retry in the for loop even have a chance of working if port numbers (ie the `webhookLocalPort`) are not available?

tests/integration/audit_config_base_suite.go
```diff
@@ -769,26 +768,17 @@ func (s *AuditConfigBaseSuite) checkWebhookAvailability() error {
...
-	}
-
-	cmd := exec.Command("kubectl", kubectlArgs...)
+	cmd := s.createPortForwardCmd(webhookService)
 	err := cmd.Start()
 	if err != nil {
 		return fmt.Errorf("couldn't start port-forward: %w", err)
 	}
 	// kubectl port-forwarding does not return but will run until interrupted
 	// We have to get ride of the port-forward at the end, because we need to create a new one for each test
```
joelddiaz, line 777:
```suggestion
	// We have to get rid of the port-forward at the end, because we need to create a new one for each test
```

tests/integration/audit_config_base_suite.go
czunker: Normally, it would wait for the process to finish. But, because I kill it before, `Wait` should only take care of the resource cleanup: https://pkg.go.dev/os#Process.Wait

From Kill docs:
> Kill causes the Process to exit immediately.
https://pkg.go.dev/os#Process.Kill

tests/integration/audit_config_base_suite.go
czunker: Yes and no.
`port-forward` will still stick around. The next loop will retry the connect and then the kill. But the `continue` will exclude trying to start the port-forward again. So it is not trying to bind to the same port.

I didn't want to add another retry loop for the stop. That look even more complicated. I'm open to other ideas...

tests/integration/audit_config_base_suite.go
czunker: Fixed

joelddiaz (approved):

PR #661, opened 2022-10-31
⭐️ namespace globbing

joelddiaz:


joelddiaz:
For the globbing to work for the CronJob-triggered scans, will need an updated Mondoo client with https://github.com/mondoohq/cnquery/pull/412 compiled in.

imilchev (approved):
LGTM. I guess we cannot merge it until a client release is done. Otherwise we will have a non-working operator feature

PR #669, opened 2022-11-04
🧹 Regularly clean up edge assets and projects

### LINKED ISSUE #659: Clean up edge assets in Mondoo
**Describe the bug**
Now that we use Mondoo edge for the edge integration tests (https://github.com/mondoohq/mondoo-operator/pull/612) we also need to clean up the assets in edge.

We already have this in place for prod: https://github.com/mondoohq/mondoo-operator/blob/main/.github/workflows/cleanup-mondoo-space.yaml

**To Reproduce**
1. Check edge console
2. Run edge integration tests
3. See assets pile up over time

**Expected behavior**
Assets should be cleaned up regularly.

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

czunker:
Edge integration tests report to Mondoo edge. So we need to cleanup there also.

Fixes #659

### DIFF
--- a/.github/workflows/cleanup-mondoo-space.yaml
+++ b/.github/workflows/cleanup-mondoo-space.yaml
@@ -2,6 +2,12 @@ name: "Cleanup Mondoo space"
 on:
   schedule:
   - cron: "11 3 * * *" # 3:11am UTC
+  workflow_dispatch:
+    secrets:
+      MONDOO_CLIENT_EDITOR:
+        required: true
+      MONDOO_CLIENT_EDITOR_EDGE:
+        required: true
 
 jobs:
   cleanup-prod:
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-mondoo-space.yaml
```diff
@@ -19,6 +25,29 @@ jobs:
           args: status
 
       - name: Cleanup Mondoo space "GitHub Actions Mondoo Operator"
+        if: success() || failure()
+        run: |
+          mondoo --config ${{ env.MONDOO_CONFIG_PATH }} status
+          ./scripts/cleanup_operator_assets.sh
```
imilchev, line 31: btw can't we use the garbage collection API for cleaning up instead of the script?

czunker: Perhaps and no.

The GC API might work for the assets, but does currently not include the CI/CD projects. For the assets, I'm not completely sure. We would need the Scan API to be still there when we clean up. I think we clean up the cluster at the end of each test.

imilchev (approved):

PR #688, opened 2022-11-16
🐛 Skip server-side applies without real change

### LINKED ISSUE #685: Webhook creates to much CI/CD assets on GKE/AKS
**Describe the bug**
image

The webhook nearly constantly creates new CI/CD assets as seen in the screenshot above.
It looks like the assets aren't created from real changes to the k8s objects.

On first sight, it looks like only DaemonSets are affected.

**To Reproduce**
1. Deploy operator to AKS/GKE cluster.
2. Enable webhook.
3. Wait a short time.

**Expected behavior**
Webhook should only create CI/CD assets on changes to the k8s objects.

czunker:
GKE, AKS, ... regularly (every 30s-60s) apply their addons. This creates update events which are send to the webhook, which creates CI/CD assets in the backend. But these updates do not contain real changes, only the `resourceVersion` changed. So, we skip them.

This is a known bug in older k8s versions: https://github.com/kubernetes/kubernetes/issues/95460

Fixes #685

### DIFF
--- a/pkg/webhooks/handler/webhook.go
+++ b/pkg/webhooks/handler/webhook.go
@@ -3,6 +3,7 @@ package webhookhandler
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"google.golang.org/protobuf/types/known/structpb"
 	admissionv1 "k8s.io/api/admission/v1"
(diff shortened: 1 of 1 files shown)


czunker:
For more background and details see the linked issue.

joelddiaz (approved):
LGTM.

This does make me think about whether we should also ignore update events to an object's Status (do those updates trigger an Update event for the webhook?). Anyway, that is work for another day. This definitely solves the noise about the resourceVersion field.

PR #479, opened 2022-08-01
🧹 Unified logging style for operator, cronjobs and admission controller

imilchev:
Unify logging format for the operator, cronjobs and admission controller. It looks like this now:
```
2022-08-01T10:08:09Z    INFO    webhook setting up manager
2022-08-01T10:08:09Z    INFO    controller-runtime.metrics      Metrics server is starting to listen    {"addr": ":8080"}
2022-08-01T10:08:09Z    INFO    webhook setting up webhook server       {"version": "sha256-1cfc847f62242675f857d25a6c8e5c38488dfe07.sig"}
2022-08-01T10:08:09Z    INFO    webhook registering webhooks to the webhook server
2022-08-01T10:08:09Z    INFO    controller-runtime.webhook      Registering webhook     {"path": "/validate-k8s-mondoo-com"}
2022-08-01T10:08:09Z    INFO    webhook starting manager
2022-08-01T10:08:09Z    INFO    controller-runtime.webhook.webhooks     Starting webhook server
2022-08-01T10:08:09Z    INFO    Starting server {"path": "/metrics", "kind": "metrics", "addr": "[::]:8080"}
2022-08-01T10:08:09Z    INFO    Starting server {"kind": "health probe", "addr": "[::]:8081"}
2022-08-01T10:08:09Z    INFO    controller-runtime.certwatcher  Updated current TLS certificate
2022-08-01T10:08:09Z    INFO    controller-runtime.webhook      Serving webhook server  {"host": "", "port": 9443}
2022-08-01T10:08:09Z    INFO    controller-runtime.certwatcher  Starting certificate watcher
```

Closes #478 

Signed-off-by: Ivan Milchev <ivan@mondoo.com>

PR #689, opened 2022-11-17
🐛 Do not show stack traces on error

### LINKED ISSUE #684: Webhook show error log and stack trace
**Describe the bug**
When the quota is reached and the webhook receives and error, it prints out a log message and a stack trace.
The trace isn't needed.

**To Reproduce**
Hard to reproduce. Perhaps is scaled down Scan API would trigger the same problem?
I saw this in a customer setup.

**Expected behavior**
I would only expect the log message. It provides enough information.

czunker:
This setting will only show stack traces on panic.

Fixes #684

### DIFF
--- a/pkg/utils/logger/logger.go
+++ b/pkg/utils/logger/logger.go
@@ -8,8 +8,9 @@ import (
 
 func NewLogger() logr.Logger {
 	opts := zap.Options{
-		Development: true,
-		TimeEncoder: zapcore.RFC3339TimeEncoder,
+		Development:     true,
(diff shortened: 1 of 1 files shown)


czunker:
As I understand https://github.com/mondoohq/mondoo-operator/pull/479#discussion_r934560355, it is only needed here to be active everywhere.

czunker:
https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.0/pkg/log/zap#Options

imilchev (approved):

PR #694, opened 2022-11-22
⭐️ Add cnspec feature flag

imilchev:
Integrate cnspec into the operator. For now I added a feature flag that allows us to switch between mondoo and cnspec so we can easily test the cnspec integration without bothering other users:

- [x] add feature flag
- [x] get proper cnspec container released that includes the scan API
- [x] validate admission controller works
- [x] validate nodes scanning works
- [x] validate cluster scanning works
- [x] extend integration tests to verify the cnspec flow


czunker (approved):
LGTM

Local tests also worked.

PR #700, opened 2022-11-25
🐛 Fix nightly build and set operator commit in binary

imilchev:
The cloud tests were broken because the version set in the `mondoo-operator` binary was `nightly` but the released container image is `main`. Due to this mismatch a very old operator binary was used with the latest mondoo client which is incompatible.

I also added the commit sha to the binary to be able to exactly pinpoint which version we are running (because`main` doesn't say much).


czunker (approved):
LGTM

PR #706, opened 2022-11-29
🧹 Make sure port-forward is really restarted

czunker:


joelddiaz (approved):
I assume this is just for extra debugging output, b/c I see no logic/behavior changes. 👍

PR #710, opened 2022-11-29
Update our spellcheck config + versions

tas50:
Use the latest in both actions and use all our patterns

Signed-off-by: Tim Smith <tsmith84@gmail.com>

### DIFF
--- a/.github/workflows/spell-check.yaml
+++ b/.github/workflows/spell-check.yaml
@@ -3,9 +3,7 @@ name: Spell Checking
 
 on:
   pull_request:
-  push:
-    branches:
-      - main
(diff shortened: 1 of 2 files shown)


.github/workflows/spell-check.yaml
```diff
@@ -49,7 +47,7 @@ jobs:
     if: (success() || failure()) && needs.spelling.outputs.followup
     steps:
       - name: comment
-        uses: check-spelling/check-spelling@v0.0.20
+        uses: check-spelling/check-spelling@v0.0.21
```
czunker, line 50: Good catch. Thanks!

czunker (approved):

PR #715, opened 2022-12-02
⭐️ Extended operator status reporting

### LINKED ISSUE #713: Operator status update needs to include boolean for subsystems being enabled
Our current status API is going to require us to bake a bunch of logic into the frontend to determine if a subsystem (image scanning, workload scanning, etc) was enabled or not. Our current data looks like this:

```json
                "messages": [
                    {
                        "message": "Kubernetes Resources Scanning is Available",
                        "status": "INFO",
                        "identifier": "k8s-resources-scanning",
                        "__typename": "IntegrationMessage"
                    },
                    {
                        "message": "Node Scanning is available",
                        "status": "INFO",
                        "identifier": "node-scanning",
                        "__typename": "IntegrationMessage"
                    },
                    {
                        "message": "No status reported yet",
                        "status": "ERROR",
                        "identifier": "admission-controller",
                        "__typename": "IntegrationMessage"
                    },
                    {
                        "message": "ScanAPI controller is available",
                        "status": "INFO",
                        "identifier": "scan-api",
                        "__typename": "IntegrationMessage"
                    }
                ],
```

The `message` field includes can be inspected to determine if the subsystem is enabled in the config, but it gets tricky since INFO vs. ERROR status changes that field. We need to expose if this service is enabled in the operator config in the cluster or not. This will help us keep logic out of the frontend.

cc: @vjeffrey @SEAjamieD

imilchev:
I addressed the problem described in #713. I discussed the changes with @preslavgerchev and we concluded that it's best to not add the booleans to the messages but to actually include them in the `LastState` object that we send upstream.

I know yesterday we discussed defining the enumeration for the different types of scans in nexus but now that I think of it I don't think it makes sense. It seems like the operator should define what it can or cannot do so the possible message identifiers are defined here.

While looking at this I also noticed that we didn't have a message for the container image scanning, so I added that now.

Fixes #713

### DIFF
--- a/controllers/mondooauditconfig_controller_test.go
+++ b/controllers/mondooauditconfig_controller_test.go
@@ -244,6 +244,11 @@ func TestTokenRegistration(t *testing.T) {
 								Status:     mondooclient.MessageStatus_MESSAGE_INFO,
 							},
 							{
+								Message:    "Container image scanning is disabled",
+								Identifier: status.ContainerImageScanningIdentifier,
+								Status:     mondooclient.MessageStatus_MESSAGE_INFO,
+							},
+							{
 								Message:    "Node scanning is disabled",
 								Identifier: status.NodeScanningIdentifier,
 								Status:     mondooclient.MessageStatus_MESSAGE_INFO,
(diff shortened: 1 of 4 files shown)


vjeffrey (approved):
thank you @imilchev !

PR #719, opened 2022-12-08
🧹 Use newer k8s versions for integration tests

imilchev:
There has been several new(er) version of k8s since we last updated our integration tests. I did that now and also dropped 1.22 since that is no longer actively supported and in a couple of months it will be EOL.
image


czunker (approved):
LGTM

PR #727, opened 2022-12-13
🧹 Complete E2E test suite

imilchev:
Migrates all tests to E2E approach where assets are verified for upstream. While doing that I found a couple of bugs that are also fixed in this PR:
- namespace filtering for container image scanning was never implemented
- it is possible to block on lack of permissions when installing the operator. If we try to use a newly created service account right away, sometimes upstream tells us we have no permissions. For now I added a 10s sleep between creating the SA and further deploying the rest of the components


tests/integration/audit_config_base_suite.go
```diff
@@ -95,8 +125,26 @@ func (s *AuditConfigBaseSuite) AfterTest(suiteName, testName string) {
...
+
+		containerScanListOpts := &client.ListOptions{Namespace: s.auditConfig.Namespace, LabelSelector: labels.SelectorFromSet(container_image.CronJobLabels(s.auditConfig))}
+		err = s.testCluster.K8sHelper.EnsureNoPodsPresent(containerScanListOpts)
+		s.NoErrorf(err, "Failed to wait for container scan pods to be gone")
+
+		nodeScanListOpts := &client.ListOptions{Namespace: s.auditConfig.Namespace, LabelSelector: labels.SelectorFromSet(nodes.CronJobLabels(s.auditConfig))}
+		err = s.testCluster.K8sHelper.EnsureNoPodsPresent(nodeScanListOpts)
+		s.NoErrorf(err, "Failed to wait for node scan pods to be gone")
+
 		// not sure why the above list does not work. It returns zero deployments. So, first a plain sleep to stabilize the test.
```
czunker, line 140: Looks like this comment is no longer valid.

tests/integration/audit_config_base_suite.go
```diff
@@ -884,3 +1051,23 @@ func (s *AuditConfigBaseSuite) createPortForwardCmd(webhookService *corev1.Servi
 
 	return exec.Command("kubectl", kubectlArgs...)
 }
+
+func (s *AuditConfigBaseSuite) AssetsNotUnscored(assets []assets.AssetWithScore) {
+	for _, asset := range assets {
+		// We don't score scratch containers at the moment so they are always unscored.
+		// We don't have policies for a cluster asset enabled at the moment so they are always unscored.
```
czunker, line 1058: Is this something we should enable for these tests?

tests/integration/audit_config_test.go
```diff
@@ -18,41 +18,41 @@ type AuditConfigSuite struct {
...
 }
 
 func (s *AuditConfigSuite) TestReconcile_Containers() {
-	auditConfig := utils.DefaultAuditConfigMinimal(s.testCluster.Settings.Namespace, false, true, false, false, false)
+	auditConfig := utils.DefaultAuditConfigMinimal(s.testCluster.Settings.Namespace, false, true, false, false)
 
-	// The mondoo-operator namespace is using local images which cnspec won't be able to download
-	auditConfig.Spec.Filtering.Namespaces.Exclude = []string{"mondoo-operator"}
+	// Ignore the operator namespace because we cannot scan a local image
+	auditConfig.Spec.Filtering.Namespaces.Exclude = []string{s.testCluster.Settings.Namespace, "kube-system"}
```
czunker, line 34: The comment and the code do not match for me: "operator namespace" vs. "kube-system".

czunker (commented):
That's a nice move of code:
image
Thanks @imilchev for these tests.

Only some nit picks. Will give it a try locally.

README.md
```diff
@@ -57,6 +57,16 @@ Many files (documentation, manifests, ...) are auto-generated. Before proposing
 2. Run `make generate` and `make test`.
 3. Commit the generated changes.
 
+### Running the integration tests locally
+To run the integration tests locally copy the `.env.example` file:
+```bash
+cp .env.example .env
+```
+Go to the Mondoo Platform and create an API token for an organization of choice. Add the API token to the `.env` file. Double-check that the API is set to the correct environment, then run:
```
czunker, line 65: Do we need "Editor" or "Owner" permissions?

czunker (approved):
Works:
```
--- PASS: TestAuditConfigUpgradeSuite (71.57s)
    --- PASS: TestAuditConfigUpgradeSuite/TestUpgradePreviousReleaseToLatest (45.35s)
PASS
ok  	go.mondoo.com/mondoo-operator/tests/integration	965.764s
```

Finally, we get this done. Thanks @imilchev for pushing it.

tests/integration/audit_config_base_suite.go
imilchev: we can't. The comments are valid. Scratch containers and the cluster asset are unscored at the moment. They are intentionally ignored

tests/integration/audit_config_test.go
imilchev: it's both that are ignored. Will update the comment

README.md
imilchev: Only owners can delete a space now. That's needed for the test cleanup so the owner role is required

czunker (approved):

PR #732, opened 2022-12-28
🧹 Use inventory definition from cnquery

### LINKED ISSUE #712: Remove inventory definition from repo
When we implemented the more complicated Mondoo use-cases at a point we copied the inventory definition to this repo (see [here](https://github.com/mondoohq/mondoo-operator/blob/main/pkg/inventory/inventory.go)). This is now part of the open-sourced cnquery. We can import the inventory definition from [here](https://github.com/mondoohq/cnquery/tree/main/motor/inventory) and remove the copied file from this repo

imilchev:
Fixes #712


joelddiaz (approved):
Nice

PR #733, opened 2023-01-04
🧹 Update dependencies

czunker:


config/crd/bases/k8s.mondoo.com_mondooauditconfigs.yaml
```diff
@@ -149,6 +149,25 @@ spec:
                     description: ResourceRequirements describes the compute resource
                       requirements.
                     properties:
+                      claims:
```
czunker, line 152: This is new in v1.26: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#resourcerequirements-v1-core

joelddiaz: Are the test failures due to older versions of k8s not knowing what to do with this new field?

czunker: Strangely, the same error happened before, when `claims` was missing. I have to dig a bit deeper to find out what's happening here.

czunker:
Reverted the 0.26.0 updates of the k8s dependencies. They aren't supported in the operator-sdk: https://github.com/operator-framework/operator-sdk/blob/master/go.mod#L41

Also reverted the controller-runtime. There is some back and forth in the releases about how to handle the generation which failed our tests: https://github.com/kubernetes-sigs/controller-tools/releases

Perhaps, with the next operator-sdk and controller release, we can upgrade to 0.26.0.

joelddiaz (approved):

PR #738, opened 2023-01-20
⭐️ report status for namespace filtering

joelddiaz:
namespace filtering is a bit different than the others...so the status is always "healthy", and as part of the report we send the currently-applied list of allow and deny list.

Then server-side the code can determine whether the currently-applied config matches the server-side desired config.


controllers/status/operator_status.go
```diff
@@ -151,6 +152,15 @@ func ReportStatusRequestFromAuditConfig(
 		messages[4].Message = "Scan API is disabled"
 	}
 
+	// Namespace filtering status
+	messages[5].Identifier = NamespaceFilteringIdentifier
```
imilchev, line 156: I think you don't need messages for this at all. The current settings of the operator are actually stored in the `LastState` object. You can add that there

imilchev: Check line 178... You can store the current filters there and remove the messages completely. The messages are only for reporting status, not configuration

imilchev (approved):
LGTM

PR #757, opened 2023-03-06
🧹 Discover only container images when scanning containers

imilchev:
Currently, all workloads are discovered when scanning container images. This works but makes the container image scan unnecessarily slower since we are scanning more assets than we are supposed to. This PR makes sure we discover only the container images when the container image scan job is running


czunker (approved):
Nice. Thanks @imilchev

PR #762, opened 2023-03-10
🐛 fix e2e tests

imilchev:
This PR fixes the e2e tests. It also requires updating `client-go` and `controller-runtime` modules since they are updated in `cnquery` on which we depend.

I also decided to commit the generated files from ranger that are needed for our tests. This speeds up our pipeline and makes the golangci-lint cache work better.


.github/workflows/lint.yaml
```diff
@@ -16,17 +16,11 @@ jobs:
...
         run: cat ".github/env" >> $GITHUB_ENV
-      - name: Install Protoc
-        run: |
-          sudo apt install -y gcc
-          make prep/ci/protoc
-          echo "${{ github.workspace }}/tools/bin" >> $GITHUB_PATH
       - uses: actions/setup-go@v3
         with:
           go-version: "${{ env.golang-version }}"
-      - run: make generate
```
czunker, line 27: What changed here, that it no longer works inside the workflow?

tests/framework/nexus/space.go
```diff
@@ -52,16 +53,18 @@ func (s *Space) ListAssetsWithScores(ctx context.Context, integrationMrn string)
...
-	if err != nil {
-		return nil, err
-	}
-
 	assetScores := make([]AssetWithScore, len(assetsPage.List))
 	for i := range assetsPage.List {
 		asset := assetsPage.List[i]
-		assetScores[i] = AssetWithScore{Asset: asset, Score: scores.Scores[asset.Mrn]}
+
+		// TODO: replace this call with GetScore(ctx, &cnspec.EntityScoreReq{EntityMrn: asset.Mrn, ScoreMrn: asset.Mrn}) once nexus is released
```
czunker, line 60: Should we already fix this now?

go.mod
```diff
@@ -9,203 +9,394 @@ require (
...
-	go.mondoo.com/cnquery v0.0.0-20221130120105-2ad23d14e576
+	github.com/stretchr/testify v1.8.2
+	go.mondoo.com/cnquery v0.0.0-20230310083103-a4ce8039521e
+	go.mondoo.com/cnspec v0.0.0-20230310120719-f285d4252c94
 	go.mondoo.com/ranger-rpc v0.5.1-0.20220923135836-9e7732899d34
-	k8s.io/api v0.25.5
-	k8s.io/apimachinery v0.25.5
-	k8s.io/client-go v0.25.5
-	sigs.k8s.io/controller-runtime v0.13.1
+	k8s.io/api v0.26.2
```
czunker, line 16: Nice to see this works now.

Last time I tried, it was still broken for operators: https://github.com/mondoohq/mondoo-operator/pull/733

.github/workflows/lint.yaml
imilchev: `make generate` pulls go dependencies and golangci-lint has its own cache of dependencies that it tries to restore. We run into conflicts while doing that. There were 2 options:
- disable golangci-lint cache
- do not do `make generate` before linting

I decided to go for the second option since that means I can also speed up the workflows a bit such that we don't need to run that command for every single task we execute for the code. The downside is that we need to commit the generated files but I think that's fine. They are quite isolated in the test packages, so I think the tradeoff is good

tests/framework/nexus/space.go
imilchev: Will fix in a couple of days when we have a back-end release. This approach works too but the other call is just a bit more efficient

czunker (approved):
Thanks for fixing this @imilchev

PR #764, opened 2023-03-14
🐛  container image gc

joelddiaz:
Container image assets are stored as "docker-image"

Make sure we use the right string when trying to garbage collect old
ones.


imilchev (approved):
LGTM! Thanks!

PR #784, opened 2023-04-17
🐛 run node scanning in privileged container for openshift

imilchev:
When scanning Openshift nodes that run RHCOS it is required to run the scan in a privileged container since otherwise we have no access to `/proc`

### DIFF
--- a/controllers/nodes/resources.go
+++ b/controllers/nodes/resources.go
@@ -40,7 +40,7 @@ const (
 	ignoreAnnotationValue = "ignore"
 )
 
-func CronJob(image string, node corev1.Node, m v1alpha2.MondooAuditConfig) *batchv1.CronJob {
+func CronJob(image string, node corev1.Node, m v1alpha2.MondooAuditConfig, isOpenshift bool) *batchv1.CronJob {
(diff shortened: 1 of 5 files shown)


controllers/nodes/resources.go
```diff
@@ -104,7 +104,9 @@ func CronJob(image string, node corev1.Node, m v1alpha2.MondooAuditConfig) *batc
 												"ALL",
 											},
 										},
-										Privileged: pointer.Bool(false),
+										// RHCOS requires to run as privileged to properly do node scanning. If the container
+										// is not privileged, then we have no access to /proc.
+										Privileged: pointer.Bool(isOpenshift),
```
chris-rock, line 109: In future version we probably want to rename the flag to `isPriviledged`

chris-rock (approved):

controllers/nodes/resources.go
imilchev: I think on the long term that should be exposed as a setting in the `MondooAuditConfig`. Other OS flavours might require privileged too. We should allow tech-savvy users to set this themselves when needed

PR #786, opened 2023-04-25
🧹 Bump k8s versions

czunker:


.github/workflows/integration-tests.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        k8s-version: [v1.23.14, v1.24.8, v1.25.4]
+        k8s-version: [v1.24.12, v1.25.8, v1.26.3]
```
imilchev, line 33: perhaps we should start testing with 1.27 as well

.github/workflows/cloud-tests.yaml
```diff
@@ -132,7 +132,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        k8s-version: ["1.22", "1.23", "1.24"]
+        k8s-version: ["1.22", "1.23", "1.24", "1.25", "1.26"]
```
imilchev, line 135: 1.22 and 1.23 are EOL. Not sure if we want to still test for those

.github/workflows/integration-tests.yaml
czunker: That's not yet supported.

.github/workflows/cloud-tests.yaml
czunker: We can drop 1.22, it will be EOL for EKS in about a month. But 1.23 will still be supported until October this year (EKS only). Looks like AWS has a longer support policy.

.github/workflows/integration-tests.yaml
imilchev: It's not on the officially supported list but it is supported. We can still deploy 1.27

imilchev:
Should also update the READMe since it specifies the versions agains which we test

.github/workflows/integration-tests.yaml
czunker: I added v1.27.1

czunker:
> Should also update the READMe since it specifies the versions agains which we test

Fixed

imilchev (approved):
LGTM

PR #789, opened 2023-05-01
⭐️ node garbage collection

imilchev:
**The feature is behind a feature flag** `FEATURE_ENABLE_NODE_GC`

Implement node garbage collection. We now have a new `CronJob` that gets deployed and which performs the garbage collection. It is executed once every 2 hours which and deletes nodes that have not been scanned in the past 2h.


chris-rock (dismissed):
Thank you @imilchev

PR #796, opened 2023-05-10
🧹 set http timeout for cnspec api to 30min

imilchev:
Just to make sure that we can properly garbage collect large spaces


czunker (approved):
:+1:

PR #797, opened 2023-05-10
🧹 remove node garbage collect feature flag and extend tests

imilchev:
Removed the node garbage collect flag and it is now enabled by default

I also extended the unit tests and integration tests to check that the new cronjob is properly created.

### DIFF
--- a/tests/integration/audit_config_base_suite.go
+++ b/tests/integration/audit_config_base_suite.go
@@ -175,7 +175,7 @@ func (s *AuditConfigBaseSuite) testMondooAuditConfigNodes(auditConfig mondoov2.M
 	cronJobs := &batchv1.CronJobList{}
 	cronJobLabels := nodes.CronJobLabels(auditConfig)
 
-	// Lits only the CronJobs in the namespace of the MondooAuditConfig and only the ones that exactly match our labels.
+	// List only the CronJobs in the namespace of the MondooAuditConfig and only the ones that exactly match our labels.
 	listOpts := &client.ListOptions{Namespace: auditConfig.Namespace, LabelSelector: labels.SelectorFromSet(cronJobLabels)}
 
 	nodeList := &corev1.NodeList{}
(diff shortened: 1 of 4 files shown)


tests/integration/audit_config_base_suite.go
```diff
@@ -215,6 +215,25 @@ func (s *AuditConfigBaseSuite) testMondooAuditConfigNodes(auditConfig mondoov2.M
...
+
+	// Verify the amount of CronJobs created is 1
+	err = s.testCluster.K8sHelper.ExecuteWithRetries(func() (bool, error) {
+		s.NoError(s.testCluster.K8sHelper.Clientset.List(s.ctx, gcCronJobs, gcListOpts))
+		if 1 == len(cronJobs.Items) {
+			return true, nil
+		}
+		return false, nil
+	})
+	s.NoErrorf(
```
imilchev, line 233: Since this runs bi-hourly we cannot really test for a successful run. That's why I left it out

tests/integration/audit_config_base_suite.go
```diff
@@ -215,6 +215,25 @@ func (s *AuditConfigBaseSuite) testMondooAuditConfigNodes(auditConfig mondoov2.M
 		s.NoErrorf(err, "Couldn't find NodeScan Pod for node "+node.Name+" in Podlist of the MondooAuditConfig Status")
 	}
 
+	// Verify the garbage collect cron job
+	gcCronJobs := &batchv1.CronJobList{}
+	gcCronJobLabels := nodes.GarbageCollectCronJobLabels(auditConfig)
+
+	// Lits only the CronJobs in the namespace of the MondooAuditConfig and only the ones that exactly match our labels.
```
czunker, line 222: nit:
```suggestion
	// List only the CronJobs in the namespace of the MondooAuditConfig and only the ones that exactly match our labels.
```

czunker (dismissed):
LGTM. Thanks @imilchev

czunker (approved):

PR #801, opened 2023-05-17
🧹 Disable cleanup workflow

czunker:
Only disable for now. Let's see how the new automatic cleanup works out. We can remove this completly in a couple of days.


imilchev (approved):
I completely missed that PR. Sorry!

Issue #213, opened 2022-03-22
Expose custom metrics


joelddiaz:
Set up the pieces to allow reporting custom metrics. https://book.kubebuilder.io/reference/metrics.html

Start by exposing a count of the number of MondooAuditConfig resources as the first custom metric. Since this is not something that can easily be calculated during a Reconcile() loop, it may make more sense to set up a controller that just periodically runs to calculate the metrics to be reported.

 https://github.com/openshift/cloud-credential-operator/blob/master/pkg/operator/metrics/ can be used as an example of setting up a controller that doesn't watch for a specific resource, but just runs at a set interval.

PR #812, opened 2023-05-28
:sparkles: add custom metrics

mariuskimmina:
#213 

Currently, this only adds one metrics `mondoo_audit_config` which shows the number of mondoAuditConfigs.

### DIFF
--- a/cmd/mondoo-operator/operator/cmd.go
+++ b/cmd/mondoo-operator/operator/cmd.go
@@ -26,6 +26,7 @@ import (
 	k8sv1alpha2 "go.mondoo.com/mondoo-operator/api/v1alpha2"
 	"go.mondoo.com/mondoo-operator/controllers"
 	"go.mondoo.com/mondoo-operator/controllers/integration"
+	"go.mondoo.com/mondoo-operator/controllers/metrics"
 	"go.mondoo.com/mondoo-operator/controllers/resource_monitor"
 	"go.mondoo.com/mondoo-operator/controllers/resource_monitor/scan_api_store"
 	"go.mondoo.com/mondoo-operator/controllers/status"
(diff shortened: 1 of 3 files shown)


mariuskimmina:
Set as draft MR so we can discuss which other metrics should be exposed

czunker:
Hi @mariuskimmina ,

thanks for your PR.

The idea behind the issue was to get to the next capability level for the operator:
image

As you can see [here](https://artifacthub.io/packages/olm/community-operators/mondoo-operator), we are currently at level 3. We do not have any metrics we currently need. The amount of configs is a nice first indicator.

We appreciate your contribution and think this is the right direction going forward. It would be nice if you could add tests for the metrics.

mariuskimmina:
thanks for the insight, I should be able to add tests sometime next week

czunker:
@mariuskimmina Thanks for adding the tests.

Strangely, now the linter step fails, which seems unrelated to your PR:
```
Error: /usr/bin/tar: ../../../go/pkg/mod/go.uber.org/multierr@v1.9.0/error.go: Cannot open: File exists
```
And a lot more like these. I need to look into this.

czunker:
There are some issues, e.g., https://github.com/golangci/golangci-lint-action/issues/244 and https://github.com/golangci/golangci-lint-action/issues/135 which mention this problem. Perhaps the solution is just disabling some conflicting caching...

czunker:
Strange, this job shows the same errors but still passes: https://github.com/mondoohq/mondoo-operator/actions/runs/5390664672/jobs/9833822310

mariuskimmina:
Linter issues have been fixed @czunker

mariuskimmina:
The failing test seem to be due to issues pulling images: `trying and failing to pull image`

czunker (approved):
Thanks for this contribution @mariuskimmina

PR #816, opened 2023-05-31
Bump cert-manager to v1.12.1

czunker:
This also required some other version bumps with possible breaking changes.


czunker:
After https://github.com/mondoohq/mondoo-operator/pull/873

imilchev (approved):

PR #828, opened 2023-06-30
🐛 Fix tar errors during golangci-lint execution

czunker:
The errors polutew the logs and make it hard to find real problems

For some background, see https://github.com/mondoohq/mondoo-operator/pull/812

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -22,5 +22,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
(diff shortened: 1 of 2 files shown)


czunker:
That's actually the official recommendation: https://github.com/golangci/golangci-lint-action/issues/244#issuecomment-1052190775

Perhaps, we could also try out removing the setup-go step.

czunker:
It does not look like disabling the cache has a big influence on the performance. Just as an alternative: https://github.com/mondoohq/mondoo-operator/pull/829

PR #829, opened 2023-07-03
🐛 Fix tar errors during golangci-lint execution

czunker:
The errors polute the logs and make it hard to find real problems

For some background, see https://github.com/mondoohq/mondoo-operator/pull/812


czunker:
This is the alternative to #828 Here, we lose the ability to choose the go version for the linting step.

imilchev (approved):
I think I prefer this approach

PR #844, opened 2023-08-21
🐛 Enable Scan API also for node scanning

czunker:
The node GC depends on the Scan API. Because of that we also need to create the Scan API when only node scanning is active.

Fixes https://github.com/mondoohq/server/issues/6267

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -405,10 +405,13 @@ prep/tools/ranger:
 
 prep/repos:
 	test -x cnquery || git clone https://github.com/mondoohq/cnquery.git
+# workaround for now. Needs to be removed as we move to v9
+	cd cnquery && git fetch && git checkout v8
 	test -x cnspec || git clone https://github.com/mondoohq/cnspec.git
 
 prep/repos/update: prep/repos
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Thank you @czunker

Makefile
```diff
@@ -405,10 +405,13 @@ prep/tools/ranger:
 
 prep/repos:
 	test -x cnquery || git clone https://github.com/mondoohq/cnquery.git
+# workaround for now. Needs to be removed as we move to v9
```
chris-rock, line 408: nit: whitespace

Makefile
```diff
@@ -405,10 +405,13 @@ prep/tools/ranger:
...
+	cd cnquery && git fetch && git checkout v8
 	test -x cnspec || git clone https://github.com/mondoohq/cnspec.git
 
 prep/repos/update: prep/repos
-	cd cnquery; git checkout main && git pull; cd -;
+# workaround for now. Needs to be removed as we move to v9
```
chris-rock, line 413: nit: whitespace

PR #864, opened 2023-09-25
🚀 1.16.0

imilchev:


charts/mondoo-operator/templates/deployment.yaml
```diff
@@ -32,10 +32,12 @@ spec:
   replicas: {{ .Values.controllerManager.replicas }}
   selector:
     matchLabels:
+      app.kubernetes.io/name: mondoo-operator
```
czunker, line 35: IIRC, that shouldn't be there. Otherwise we will have the label twice, which results in problems with tools like Flux CD.

charts/mondoo-operator/templates/manager-config.yaml
```diff
@@ -1,6 +1,3 @@
-# Copyright (c) Mondoo, Inc.
```
czunker, line 1: Should we re-add these after the generation?

imilchev: I don't know yet. Let's see what the licence checker will say. I think we should ignore the yaml files for that check though...

charts/mondoo-operator/templates/deployment.yaml
imilchev: shouldn't these not be showing up then? i thought this was fixed somewhere so we don't get those anymore. Do I need to remove them manually or did something else break?

czunker: These lines should remove the tags:
https://github.com/mondoohq/mondoo-operator/blob/main/Makefile#L363-L365

imilchev: yes, that was it. I had to use `gsed` instead...

czunker (approved):
Thanks @imilchev

czunker (approved):

PR #867, opened 2023-09-27
🐛 cnspec v9 action

imilchev:
Can be triggered manually. As input we provide the version of cnspec to release and the action will build the following container flavours:
- normal
- rootless
- ubi
- ubi rootless

Run: https://github.com/mondoohq/mondoo-operator/actions/runs/6324829721


chris-rock (approved):
Thank you @imilchev

PR #874, opened 2023-10-04
🧹 make sure cnspec release works with versions containing 'v'

imilchev:
The cnspec releases always start with a leading `v`. The action that releases the pre-bundled cnspec with the required providers for the operator needs the version to have no leading `v`. After merging this we can add a step in the cnspec release flow to call this workflow and automatically build the container for every cnspec release

Successful run https://github.com/mondoohq/mondoo-operator/actions/runs/6402540882


czunker (approved):
LGTM. Thanks @imilchev

PR #883, opened 2023-10-06
🐛 release the correct cnspec container flavours

imilchev:
currently, we always use the `root` image as base. This PR makes sure we always start with a `root` image, install the providers in the system directory and then we switch the user to `101` to make it rootless. We release only `X-rootless` and `X-ubi-rootless` images since those are the only ones that the operator uses

### DIFF
--- a/.github/workflows/cnspec.yaml
+++ b/.github/workflows/cnspec.yaml
@@ -26,8 +26,6 @@ jobs:
       matrix:
         suffix:
           - ""
-          - -rootless
-          - -ubi-rootless
           - -ubi
 
     steps:
@@ -67,7 +65,7 @@ jobs:
             type=semver,pattern={{major}},value=${{ env.VERSION }}
             type=raw,value=latest
           flavor: |
-            suffix=${{ matrix.suffix }},onlatest=true
+            suffix=${{ matrix.suffix }}-rootless,onlatest=true
 
       - name: Build and push cnspec image
         id: build-and-push-operator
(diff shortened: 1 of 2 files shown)


.github/workflows/cnspec.yaml
```diff
@@ -67,16 +65,16 @@ jobs:
...
+            suffix=${{ matrix.suffix }}-rootless,onlatest=true
 
       - name: Build and push cnspec image
         id: build-and-push-operator
         uses: docker/build-push-action@v5
         with:
           context: .
           file: cnspec.Dockerfile
-          build-args: VERSION=${{ env.VERSION }}
-          platforms: linux/amd64,linux/arm64,linux/arm
```
imilchev, line 79: the UBI thing doesn't have `arm7` version and for now we don't really need it

imilchev:
successful run https://github.com/mondoohq/mondoo-operator/actions/runs/6432698630

cnspec.Dockerfile
```diff
@@ -7,4 +7,6 @@ FROM mondoo/cnspec:$VERSION
 
 RUN cnspec providers install os
 RUN cnspec providers install network
-RUN cnspec providers install k8s
\ No newline at end of file
+RUN cnspec providers install k8s
+
+USER 100:101
```
czunker, line 12: I'm not sure about this. Is the user already present? Do we need to create it before switching to it?

imilchev: no, we already do that in cnquery for the UBI image. For the non-UBI it works as well because the operator actually overrides the `mondoo` user even now to user `100`

czunker (approved):

PR #900, opened 2023-10-26
🐛 Use defined scanner image also for container scan

### LINKED ISSUE #887: Container scan image is does not follow MondooAuditConifg overrides
If we have specified a container image override in `MondooAuditConfig`, we don't use it when creating the container scan cronjob. We should grab it from `scanner.image`

czunker:
When the `MondooAuditConfig` defiens a custom image, also use it for the container scan `CronJob`.

Fixes #887


czunker:
Also tried it locally:
```
k get -n mondoo-operator mondooauditconfigs.k8s.mondoo.com mondoo-client -o yaml | grep -A3 image                                                                --
    image:
      name: nginx
      tag: 1.25-alpine
```
```
k get -n mondoo-operator cronjob.batch/mondoo-client-containers-scan -o yaml | grep image                                                                     
            image: nginx:1.25-alpine
```

imilchev (approved):
nice that we also have a test for this! :shipit:

PR #903, opened 2023-10-27
🐛 Fix asset garbage collection

### LINKED ISSUE #893: Kubernetes resources garbage collection is not working in v9
**Describe the bug**
Running the operator with the `FEATURE_ENABLE_V9` flag enabled. I can see that my kubernetes resource assets aren't getting garbage collected by the operator. I don't see any errors in the `k8s-scan` cronjob.

**To Reproduce**

1. Install the operator on a cluster
2. Enable v9
3. Create a few workloads and let the operator scan then with a cluster scan
4. Delete the workloads
5. Wait for a couple of hours and see that they are still in the UI

**Expected behavior**
Old kubernetes resources should disappear after 2h

**Screenshots or CLI Output**
image

czunker:
Fix runtime and start separate jobs for objects and container images.

Fixes #893

### DIFF
--- a/cmd/mondoo-operator/garbage_collect/cmd.go
+++ b/cmd/mondoo-operator/garbage_collect/cmd.go
@@ -109,7 +109,7 @@ func GarbageCollectCmd(ctx context.Context, client scanapiclient.ScanApiClient, 
 
 	if platformRuntime != "" {
 		switch platformRuntime {
-		case "k8s", "docker-image":
+		case "k8s-cluster", "docker-image":
 			gcOpts.PlatformRuntime = platformRuntime
 		default:
 			return fmt.Errorf("no matching platform runtime found for (%s)", platformRuntime)
--- a/cmd/mondoo-operator/k8s_scan/cmd.go
+++ b/cmd/mondoo-operator/k8s_scan/cmd.go
@@ -96,15 +96,23 @@ func init() {
 
 		// If scanning successful, now attempt some cleanup of older assets
 		if *setManagedBy != "" && *cleanupOlderThan != "" {
-			platformRuntime := "k8s"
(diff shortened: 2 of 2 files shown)


czunker:
This also requires https://github.com/mondoohq/cnspec/pull/895 to fix https://github.com/mondoohq/mondoo-operator/issues/893

cmd/mondoo-operator/garbage_collect/cmd.go
```diff
@@ -109,7 +109,7 @@ func GarbageCollectCmd(ctx context.Context, client scanapiclient.ScanApiClient,
 
 	if platformRuntime != "" {
 		switch platformRuntime {
-		case "k8s", "docker-image":
+		case "k8s-cluster", "docker-image":
```
czunker, line 112: I double checked this with the v8 client. There, we already have the "k8s-cluster" runtime:
```
cnquery-8.12.0 run k8s -c "asset{ name runtime }"                                                                                                                          
→ loaded configuration from /etc/opt/mondoo/mondoo.yml using source default
→ discover related assets for 1 asset(s)
→ discovery option auto is used. This will detect the assets: cluster, cronjobs, daemonsets, deployments, ingresses, jobs, pods, replicasets, statefulsets
→ use cluster name from kube config cluster-name=minikube
→ resolved assets resolved-assets=44
asset: {
  runtime: "k8s-cluster"
  name: "K8s Cluster minikube"
}
asset: {
  runtime: "k8s-cluster"
  name: "cert-manager/cert-manager-webhook"
}
asset: {
  runtime: "k8s-cluster"
  name: "kube-system/coredns"
}
....
```

imilchev (approved):

PR #928, opened 2023-11-08
✨ Report on OOMkilled status

czunker:
In case one the scan Pods (node, container images, k8s resources, scan API) get killed due to OOM, we will add this to the status and also upstream.

This should help to better understand the health of the system.


czunker:
This is still work in progress, but for the Scan API, the `MondooAuditConfig` gets updated with OOM status:
```
- lastTransitionTime: "2023-11-08T12:06:14Z"
  lastUpdateTime: "2023-11-08T12:12:53Z"
  message: ScanAPI controller is unavailable due to OOM
  reason: ScanAPIUnvailable
  status: "True"
  type: ScanAPIDegraded
```

czunker:
The same example for a node scan CronJob:
```
affectedPods:
  - mondoo-client-node-minikube-28333730-blbw9
lastTransitionTime: "2023-11-15T04:54:06Z"
lastUpdateTime: "2023-11-15T04:54:06Z"
memoryLimit: 20Mi
message: Node Scanning is unavailable due to OOM
reason: NodeScanningUnavailable
status: "True"
type: NodeScanningDegraded
```

imilchev (approved):
LGTM! I think it would be nice to add 1 E2E test where you set a cronjob memory limit to a very low value, such that it gets OOMKilled. Then you can check the integration messages, to verify the OOM has been reported to nexus

imilchev:
there was also a section in the card about reporting mondoo-operator `OOMKilled`. Is that intentionally not part of this PR?

czunker:
> there was also a section in the card about reporting mondoo-operator `OOMKilled`. Is that intentionally not part of this PR?

Yes, I'm currently working on this in a separate PR.

czunker:
> LGTM! I think it would be nice to add 1 E2E test where you set a cronjob memory limit to a very low value, such that it gets OOMKilled. Then you can check the integration messages, to verify the OOM has been reported to nexus

I'm not sure about the e2e test. I'm afraid it will cause more trouble than help. I'll create a follow up issue for that.

PR #929, opened 2023-11-09
🐛 Fix GKE cloud test

czunker:
`GitHubSecurityLab/actions-permissions/monitor` added a MITM proxy to intercept calls. This proxy signed the GKE API cert which lead to errors:
```
Unable to connect to the server: tls: failed to verify certificate: x509: certificate signed by unknown authority
```


imilchev (approved):
LGTM

PR #931, opened 2023-11-10
🐛 Fetch more assets from upstream

czunker:
This fixes the failing OLM tests where we have more workloads than for other tests. The default number of assets fetched from upstream is 25. That wasn't enough for OLM clusters.

### DIFF
--- a/tests/framework/nexus/assets/assets.go
+++ b/tests/framework/nexus/assets/assets.go
@@ -44,7 +44,7 @@ func ListAssetsWithScores(
 					AssetType string `graphql:"asset_type"`
 				}
 			}
-		} `graphql:"assets(spaceMrn: $spaceMrn)"`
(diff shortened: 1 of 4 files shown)


tests/framework/nexus/assets/assets.go
```diff
@@ -44,7 +44,7 @@ func ListAssetsWithScores(
 					AssetType string `graphql:"asset_type"`
 				}
 			}
-		} `graphql:"assets(spaceMrn: $spaceMrn)"`
+		} `graphql:"assets(spaceMrn: $spaceMrn, first: 100)"`
```
czunker, line 47: That's the actual bugfix.

czunker:
IIRC, this will also help with some cloud tests.

imilchev (approved):

PR #937, opened 2023-11-15
📄 Add docs on how to set resources

czunker:


imilchev (approved):

imilchev:
perhaps @misterpantz wants to check this too since we are going to be sending that to customers?

misterpantz:
Thank you, @imilchev! I'm out sick today. I think what would be best is if I make a new PR with my edits to the whole document. I don't think I've ever seen this! So I'd love to take a pass when I'm healthy.

imilchev (approved):

PR #938, opened 2023-11-15
✨ Expose CronJob schedule via config

### LINKED ISSUE #888: Make `CronJob` schedule configurable
In clusters where change doesn't happen that frequently, it might not be necessary to scan every hour.

Expose the CronJob schedule via the `MondooAuditConfig`, so it can be changed according to the cluster's needs.

We could also check whether objects have actually changed since the last scan. But this might include a state that we need to store and make the overall scanning more complex. Exposing the schedule should be the easier solution.

czunker:
It is now possible to change the CronJob schedule for the different types of CronJobs via the `MondooAuditConfig`.

Fixes #888

### DIFF
--- a/controllers/nodes/resources.go
+++ b/controllers/nodes/resources.go
@@ -16,6 +16,8 @@ import (
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/yaml"
 
+	// That's the mod k8s relies on https://github.com/kubernetes/kubernetes/blob/master/go.mod#L63
+	"github.com/robfig/cron/v3"
 	"go.mondoo.com/cnquery/v9/providers-sdk/v1/inventory"
 	"go.mondoo.com/mondoo-operator/api/v1alpha2"
 	"go.mondoo.com/mondoo-operator/controllers/scanapi"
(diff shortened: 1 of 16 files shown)


controllers/nodes/resources.go
```diff
@@ -193,7 +204,7 @@ func CronJob(image string, node corev1.Node, m v1alpha2.MondooAuditConfig, isOpe
 func GarbageCollectCronJob(image, clusterUid string, m v1alpha2.MondooAuditConfig) *batchv1.CronJob {
 	ls := CronJobLabels(m)
 
-	cronTab := fmt.Sprintf("%d */2 * * *", time.Now().Add(1*time.Minute).Minute())
+	cronTab := fmt.Sprintf("%d */12 * * *", time.Now().Add(1*time.Minute).Minute())
```
czunker, line 207: I extended this a bit because I do not expect too many changes in nodes, but I didn't want to expose this. That's something users shouldn't care about.

The `--filter-older-than` has to be >= the schedule for the node scan. This expects people not to scan nodes less often than every two days.

This isn't ideal. I'm open to other ideas. Should we also add a warning in the docs?

imilchev (approved):
LGTM

PR #939, opened 2023-11-16
✨ Detect OOMkilled status for the operator itself

czunker:
This adds a Status Condition for the controller itself:
```
 - affectedPods:
    - mondoo-operator-controller-manager-7594c45888-j5xf6
  lastTransitionTime: "2023-11-16T11:22:10Z"
  lastUpdateTime: "2023-11-16T11:22:10Z"
  memoryLimit: 15Mi
  message: Mondoo Operator controller is unavailable due to OOM
  reason: MondooOperatorUnvailable
  status: "True"
  type: MondooOperatorDegraded
```


czunker:
- [ ] The revert of the status in test does not work. Needs a fix.

Hit https://github.com/kubernetes-sigs/controller-runtime/issues/2362 in the tests, but the issue provides a workaround.

controllers/metrics/metrics.go
```diff
@@ -70,7 +70,6 @@ func (mr *MetricsReconciler) Start(ctx context.Context) error {
 }
 
 func (mr *MetricsReconciler) metricsLoop() {
-	mr.log.Info("Updating metrics")
```
czunker, line 73: This one was filling up the logs. We don't have a debug level, so I removed it.

imilchev (commented):
can you add an integration test for this? should be an easily reproducible scenario.

czunker:
> can you add an integration test for this? should be an easily reproducible scenario.

Good call. The tests revealed some cases where the code wouldn't have updated the status correctly.

chris-rock (approved):
Thank you @czunker

PR #944, opened 2023-11-20
🧹 Add e2e OOM tests

czunker:
This adds e2e tests for OOM of Scan API and node scan. This covers the two different cases we currently have: Deployment and CronJob.

Fixes #940

PR #948, opened 2023-11-21
🐛 Fix CronJob schedule

czunker:
The CronJob schedule was changed with every reconcile. This PR adds the schedule during CronJob creation to the MondooAuditConfig so that it can later be re-used.

### DIFF
--- a/controllers/k8s_scan/deployment_handler_test.go
+++ b/controllers/k8s_scan/deployment_handler_test.go
@@ -73,6 +73,8 @@ func (s *DeploymentHandlerSuite) AfterTest(suiteName, testName string) {
 
 func (s *DeploymentHandlerSuite) TestReconcile_Create() {
 	d := s.createDeploymentHandler()
+	mondooAuditConfig := &s.auditConfig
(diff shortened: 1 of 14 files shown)


czunker:
The bug broke the pipeline tests in https://github.com/mondoohq/mondoo-operator/pull/944 .

The schedule was always pushed a minute and created new Jobs, changing the number of expected assets/objects.

I guess we had the bug for quite some time, but hit it now because with https://github.com/mondoohq/mondoo-operator/pull/938 we checked the `Schedule` during reconcile.

controllers/k8s_scan/deployment_handler_test.go
```diff
@@ -319,8 +327,56 @@ func (s *DeploymentHandlerSuite) TestReconcile_Disable() {
 	s.Equal(0, len(cronJobs.Items))
 }
 
+func (s *DeploymentHandlerSuite) TestReconcile_CreateWithDefaultSchedule() {
```
czunker, line 330: I added the test only here because the 61s sleep adds significant time to the tests.

But the integration tests cover every CronJob.

chris-rock (approved):
Thank you @czunker

PR #993, opened 2024-01-08
🐛 Fix upload artifact

czunker:
Fixes '(409) Conflict: an artifact with this name already exists on the workflow run'.

For background: https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -81,7 +81,7 @@ jobs:
       - uses: actions/upload-artifact@v4  # upload test results
         if: success() || failure()        # run this step even if previous step failed
         with:                             # upload a combined archive with unit and integration test results
-          name: test-results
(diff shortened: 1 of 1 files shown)


czunker:
I think this should fix the failing tests in #992 and #989

.github/workflows/integration-tests.yaml
```diff
@@ -81,7 +81,7 @@ jobs:
       - uses: actions/upload-artifact@v4  # upload test results
         if: success() || failure()        # run this step even if previous step failed
         with:                             # upload a combined archive with unit and integration test results
-          name: test-results
+          name: test-results-${{ matrix.k8s-distro }}-${{ matrix.k8s-version }}
```
imilchev, line 84: just that probably won't work. The artifact is referenced in https://github.com/mondoohq/mondoo-operator/blob/main/.github/workflows/test-report.yaml I think changing the name will break this

imilchev:
will be handled in #1026

PR #999, opened 2024-01-11
🐛 fix endless updates of cronjobs

imilchev:
2 issues here make the cronjobs be updated endlessly and never execute:
- the concurrency policy for cronjobs wasn't properly set in the update flow
- the crontab was never updated in the `MondooAuditConfig` when cronjobs just had to be updated. This meant that for existing installations of the operator, the cronjobs would never match what is in the `MondooAuditConfig`. The CRD would always have an empty `Schedule` and the cronjobs would always try to use the current time to create a schedule

### DIFF
--- a/controllers/mondooauditconfig_controller.go
+++ b/controllers/mondooauditconfig_controller.go
@@ -5,6 +5,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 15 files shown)


pkg/utils/k8s/equality.go
```diff
@@ -67,23 +67,60 @@ func AreServicesEqual(a, b corev1.Service) bool {
...
-		AreResouceRequirementsEqual(aPodSpec.Containers[0].Resources, bPodSpec.Containers[0].Resources) &&
-		AreSecurityContextsEqual(aPodSpec.Containers[0].SecurityContext, bPodSpec.Containers[0].SecurityContext) &&
-		reflect.DeepEqual(aPodSpec.Volumes, bPodSpec.Volumes) &&
-		reflect.DeepEqual(a.Spec.SuccessfulJobsHistoryLimit, b.Spec.SuccessfulJobsHistoryLimit) &&
-		reflect.DeepEqual(a.Spec.ConcurrencyPolicy, b.Spec.ConcurrencyPolicy) &&
-		a.Spec.Schedule == b.Spec.Schedule &&
-		reflect.DeepEqual(a.Spec.FailedJobsHistoryLimit, b.Spec.FailedJobsHistoryLimit) &&
-		reflect.DeepEqual(a.GetOwnerReferences(), b.GetOwnerReferences())
+
+	if len(aPodSpec.Containers) != len(bPodSpec.Containers) {
```
imilchev, line 71: the logic is the same but just split the statements so I can put a breakpoint in each and see where the check fails

controllers/mondooauditconfig_controller.go
```diff
@@ -166,6 +167,30 @@ func (r *MondooAuditConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 	}
 
+	// Set the default cron tab if none is set
+	defaultCronTab := fmt.Sprintf("%d * * * *", time.Now().Add(1*time.Minute).Minute())
```
czunker, line 171: Previously, this was different for the container scan job. https://github.com/mondoohq/mondoo-operator/blob/648c2895e771f094dcccbcf5aff06241282efd78/controllers/container_image/resources.go#L53

imilchev:
> > * the crontab was never updated in the `MondooAuditConfig` when cronjobs just had to be updated. This meant that for existing installations of the operator, the cronjobs would never match what is in the `MondooAuditConfig`. The CRD would always have an empty `Schedule` and the cronjobs would always try to use the current time to create a schedule
> 
> Can you explain a bit more about the 2nd problem? I thought I fixed it some time ago: https://github.com/mondoohq/mondoo-operator/pull/948/files

The way it is implemented now is that it only sets the crontab in the `MondooAuditConfig` when the cronjobs are created. However, for existing installations the cronjobs already exist and just need to be updated. Because they are being updated we never set the crontabs in the `MondooAuditConfig` and every minute the cronjobs are being updated with the current time

czunker (approved):
Nice! Thanks @imilchev

PR #1006, opened 2024-01-23
✨ add feature flag for using cnspec v10

imilchev:
remove the old v9 feature flag and add a feature flag for v10

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -46,9 +46,6 @@ spec:
         image: controller:latest
         imagePullPolicy: IfNotPresent
         name: manager
-        env:
-          - name: FEATURE_ENABLE_V9
-            value: "true"
         securityContext:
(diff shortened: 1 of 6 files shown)


czunker (approved):
LGTM Thanks @imilchev

config/manager/manager.yaml
```diff
@@ -46,9 +46,6 @@ spec:
         image: controller:latest
         imagePullPolicy: IfNotPresent
         name: manager
-        env:
-          - name: FEATURE_ENABLE_V9
```
czunker, line 50: After reading your Discord comment:

Could we add a test for this? I understood, it would currently fail. It's more a general question.

imilchev: we could. I would rather merge this and once we have a v10 would make sure all tests run with v10. We want to migrate everyone for v10 so there's probably no point to add tests for both v9 and v10

PR #1021, opened 2024-02-21
🧹 use prod for integration testing

imilchev:
- use prod for running the integration tests
- switch to go 1.22


chris-rock (approved):
Thank you @imilchev

PR #1027, opened 2024-02-26
✨ load operator image from operator instead of hardcoded

imilchev:
Instead of using a hard-coded image, get the operator image from the operator pod. If a user has a custom image of the operator, it will be picked up automatically by default


pkg/utils/mondoo/container_image_resolver.go
```diff
@@ -34,20 +40,35 @@ type ContainerImageResolver interface {
...
+	imageCacher          imagecache.ImageCacher
+	kubeClient           client.Client
+	operatorPodName      string
+	operatorPodNamespace string
 }
 
-func NewContainerImageResolver(isOpenShift bool) ContainerImageResolver {
+func NewContainerImageResolver(kubeClient client.Client, isOpenShift bool) ContainerImageResolver {
+	podName := os.Getenv(PodNameEnvVar)
+	if podName == "" {
```
czunker, line 57: This case feels like it shouldn't happen. Should we log an error here?

pkg/utils/mondoo/container_image_resolver.go
```diff
@@ -63,8 +84,23 @@ func (c *containerImageResolver) CnspecImage(userImage, userTag string, skipImag
...
 }
 
-func (c *containerImageResolver) MondooOperatorImage(userImage, userTag string, skipImageResolution bool) (string, error) {
-	image := userImageOrDefault(MondooOperatorImage, MondooOperatorTag, userImage, userTag)
+func (c *containerImageResolver) MondooOperatorImage(ctx context.Context, userImage, userTag string, skipImageResolution bool) (string, error) {
+	image := ""
+
+	// First try with the provided user image and tag (if any)
+	if userImage != "" || userTag != "" {
+		image = userImageOrDefault(MondooOperatorImage, MondooOperatorTag, userImage, userTag)
```
czunker, line 92: This one took me some time. First, I thought it was wrong.

Suggestions to make it perhaps more readable:
Remove the `if` around it; we already have this in the func. Remove the `OrDefault` part.

pkg/utils/mondoo/container_image_resolver.go
```diff
@@ -63,8 +84,23 @@ func (c *containerImageResolver) CnspecImage(userImage, userTag string, skipImag
...
+	}
+
+	// If still no image, then load the image from the operator pod
+	if image == "" {
+		operatorPod := &corev1.Pod{}
+		if err := c.kubeClient.Get(ctx, client.ObjectKey{Namespace: c.operatorPodNamespace, Name: c.operatorPodName}, operatorPod); err != nil {
+			return "", err
+		}
+		image = operatorPod.Spec.Containers[0].Image
```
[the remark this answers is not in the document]
imilchev: will change that

pkg/utils/mondoo/container_image_resolver.go
imilchev: it can happen if someone manually updated the operator deployment to change the image to a newer version but didn't inject the new env vars. I wanted to make sure that we keep working for such cases.

pkg/utils/mondoo/container_image_resolver.go
imilchev: we need the `if`. We only use the "old" approach if the user has specified a user image or a user tag. Only if both are unspecified we read from the pod. I can invert the if statement. Perhaps that will make it more readable

PR #1031, opened 2024-02-26
🧹 update dependencies

imilchev:


chris-rock (approved):

tests/integration/audit_config_oom_test.go
```diff
@@ -142,10 +143,20 @@ func (s *AuditConfigOOMSuite) TestOOMScanAPI() {
...
+		}
 
-	cond := mondoo.FindMondooAuditConditions(foundMondooAuditConfig.Status.Conditions, mondoov2.ScanAPIDegraded)
+		cond = mondoo.FindMondooAuditConditions(foundMondooAuditConfig.Status.Conditions, mondoov2.ScanAPIDegraded)
+		if cond != nil && strings.Contains(cond.Message, "OOM") {
+			return true, nil
+		}
+		return false, nil
+	})
+	s.Require().NoError(err, "Failed to find degraded condition")
```
czunker, line 159: We are now checking twice fr the degaded condition?

See L143-L144

imilchev: The problem with the check on L143 is that it checks for degraded condition, which doesn't need to be OOM. The check would pass for any degraded condition, which can cause the test to fail. The intention is to wait for the specific OOM message. 

I can remove the other check

czunker (approved):
Thanks @imilchev

PR #1034, opened 2024-02-27
🧹 restructure test logs for e2e tests

imilchev:
this doesn't change anything functionally but makes a more human-friendly structure of the test logs. We now store the logs for a suite in a separate folder. Inside there is a folder for each individual test:
image


czunker (approved):
Nice! Thanks @imilchev

PR #1036, opened 2024-02-27
🧹 rewrite assertions for admission e2e tests

imilchev:
attempt to re-write the assertions for the admission e2e tests such that they become more robust


tests/integration/audit_config_oom_test.go
czunker, line 1: Are these changes intended to be in this PR?

I'm just asking because of the PR title.

czunker (approved):
Thanks @imilchev

PR #1041, opened 2024-02-28
✨ add real healtchecks for the admission controller

imilchev:
makes sure we actually verify we can talk to the scan api before marking the webhook as ready


czunker (approved):
Thanks @imilchev

PR #1042, opened 2024-02-28
🧹 extend the timeout for the validating webhook

imilchev:
occasionally our integration tests fail because the scan api takes longer than 10s to respond to an admission call. Here we change the default 10s timeout to a longer one to make sure the tests succeed even on slower clusters

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -32,3 +32,5 @@ webhooks:
     - jobs
     - cronjobs
   sideEffects: None
+  # The default is 10s but on slow clusters we may take longer
+  timeout: 20s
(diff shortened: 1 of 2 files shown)


chris-rock (dismissed):

config/webhook/manifests.yaml
```diff
@@ -32,3 +32,4 @@ webhooks:
     - jobs
     - cronjobs
   sideEffects: None
+  timeout: 20s
```
chris-rock, line 35: we may want to add a comment here so that we know in the future why we did it

imilchev: i can do that

imilchev: done

chris-rock (approved):

PR #1046, opened 2024-02-29
🧹 group dependabot updates

imilchev:
i took that approach from cnquery. Hope it works here too


czunker (approved):
:+1: Thanks @imilchev

PR #1053, opened 2024-03-04
🧹 assert scored policies in e2e tests

imilchev:
Explicitly assert which policies have been scored for each asset. Useful to make sure we aren't scoring policies that shouldn't be scored for a specific asset type


czunker (approved):
Thanks @imilchev

PR #1070, opened 2024-03-26
🐛 fix private image container scans

imilchev:
makes sure private container image scans work. The issue was that the container pull secrets were overriding the mondoo folder, so the mondoo config was no longer available. That resulted in failure to communicate with upstream


czunker (dismissed):
LGTM. Thanks @imilchev

czunker (approved):

PR #1076, opened 2024-03-27
⭐️ allow using deployments for node scans

imilchev:
Allow using deployments for node scans instead of cronjob. For fully utilized clusters, cronjobs may not be able to run. Even if we can schedule the job now, there is no guarantee the node will have resources 1h from now. Using deployments ensures that once the deployment is scheduled it will always run a node scan, because the resources are already reserved.

### DIFF
--- a/api/v1alpha2/mondooauditconfig_types.go
+++ b/api/v1alpha2/mondooauditconfig_types.go
@@ -90,11 +90,23 @@ type KubernetesResources struct {
 	Schedule string `json:"schedule,omitempty"`
 }
 
+// NodeScanStyle specifies the scan style for nodes
+type NodeScanStyle string
+
+const (
+	NodeScanStyle_CronJob    NodeScanStyle = "cronjob"
+	NodeScanStyle_Deployment NodeScanStyle = "deployment"
+)
+
 type Nodes struct {
 	Enable    bool                        `json:"enable,omitempty"`
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 	// Specify a custom crontab schedule for the node scanning job. If not specified, the default schedule is used.
 	Schedule string `json:"schedule,omitempty"`
+	// Style is the style of the node scanning. The default is "cronjob" which will create a CronJob for the node scanning.
+	// +kubebuilder:validation:Enum=cronjob;deployment
+	// +kubebuilder:default=cronjob
+	Style NodeScanStyle `json:"style,omitempty"`
 }
 
 type Admission struct {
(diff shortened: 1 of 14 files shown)


controllers/nodes/resources.go
```diff
@@ -52,145 +51,263 @@ func CronJob(image string, node corev1.Node, m *v1alpha2.MondooAuditConfig, isOp
...
+	}
+	dep.Spec.Replicas = ptr.To(int32(1))
+	dep.Spec.Selector = &metav1.LabelSelector{
+		MatchLabels: labels,
+	}
+	dep.Spec.Template.Labels = labels
+	dep.Spec.Template.Annotations = map[string]string{
+		ignoreQueryAnnotationPrefix + "mondoo-kubernetes-security-pod-runasnonroot": ignoreAnnotationValue,
+	}
+	dep.Spec.Template.Spec.NodeName = node.Name
```
czunker, line 206: I don't think NodeName is the best approach for the problem we are trying to solve with the Deployments.

According to the docs, this will bypass the scheduler:
https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodename
So, when no resources are free, we end in the same situation as with CronJobs.

When we use the nodeSelector with the hostname label and a priorityClassName, the scheduler is even able to evict Pods with a lower prioClass, helping to automate the bin packing.
https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#pod-selection-for-kubelet-eviction
This would alow users to better tune this to their clusters.

Without a prioClass it could also happen that our Pods get evicted when other Pods try to run on the same node.

api/v1alpha2/mondooauditconfig_types.go
```diff
@@ -90,11 +90,28 @@ type KubernetesResources struct {
...
+	// used. Only applicable for CronJob style
 	Schedule string `json:"schedule,omitempty"`
+	// IntervalTimer is the interval (in minutes) for the node scanning. The default is "60". Only applicable for Deployment
+	// style.
+	// +kubebuilder:default=60
+	IntervalTimer int `json:"intervalTimer,omitempty"`
+	// Style is the style of the node scanning. The default is "cronjob" which will create a CronJob for the node scanning.
+	// +kubebuilder:validation:Enum=cronjob;deployment
+	// +kubebuilder:default=cronjob
+	Style NodeScanStyle `json:"style,omitempty"`
```
czunker, line 114: As explained in more detail below, we should add a `priorityClassName` option here.
We can leave it empty. The user should set this as needed.

czunker: The details are in https://github.com/mondoohq/mondoo-operator/pull/1076#discussion_r1542359370

controllers/nodes/resources.go
imilchev: I am not sure how to use a node selector by the node name. I find examples only for matching by labels. We cannot rely on node labels to run the deployment on a specific node because we aren't labeling the nodes ourselves.

czunker (approved):
Nice! Thanks @imilchev

czunker (approved):

czunker (approved):

PR #1077, opened 2024-03-27
Fix dead docs links

tas50:
These have been broken for a while

### DIFF
--- a/docs/user-manual.md
+++ b/docs/user-manual.md
@@ -35,7 +35,6 @@ This user manual describes how to install and use the Mondoo Operator.
     - [Why is there a deployment marked as unschedulable?](#why-is-there-a-deployment-marked-as-unschedulable)
     - [Why are (some of) my nodes unscored?](#why-are-some-of-my-nodes-unscored)
     - [How can I trigger a new scan?](#how-can-i-trigger-a-new-scan)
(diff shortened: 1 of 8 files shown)


docs/user-manual.md
```diff
@@ -827,7 +834,7 @@ Note: This is not possible immediately after performing the operator upgrade.
    kubectl get mondooauditconfigs.v1alpha1.k8s.mondoo.com mondoo-client -n mondoo-operator -o yaml > audit-config.yaml
    ```
 
-2. Map the old `v1alpha1` config to the new `v1alpha2` and save the new `MondooAuditConfig`. Find the mapping from `v1alpha1` to `v1alpha2` [here](../api/v1alpha1/mondooauditconfig_types.go#L155-L199).
```
czunker, line 830: I don't think the last sentence makes sense any more when we remove the `here`.

tas50: @czunker Is there a place we can continue to link folks?

czunker: I just checked the history. We removed the v1alpha1 API 2 years ago:
image

I think we can remove the complete
```
### I had a `MondooAuditConfig` in my cluster with version `v1alpha1` and now I can no longer access it. What should I do?
```

tas50: I nuked it all. Thanks for looking around

czunker (approved):
Thanks @tas50

PR #1078, opened 2024-03-28
🐛 route container resolving via http proxy

imilchev:
If a HTTP proxy is specified, make sure the container resolution traffic is routed through it too


czunker (approved):
Thanks @imilchev

PR #1079, opened 2024-03-28
🧹 increase default resource requirements

imilchev:
We have seen some issues with the current default resource requirements so we increase them


czunker (approved):
Thanks @imilchev

PR #1098, opened 2024-04-30
✨ allow setting custom env vars for container and node scans

imilchev:
Allow setting custom env vars for container and node scans. This is useful when we want to debug certain behaviour where we need to set the `DEBUG` or `MEM_DEBUG` env vars, for example


czunker (dismissed):
Thanks @imilchev

chris-rock (approved):

PR #1100, opened 2024-04-30
🧹 do not error out when running operator outside of the cluster

imilchev:
if we run the operator outside of the cluster, it tries to fetch the operator deployment inside the cluster, which doesn't exist. In that case we cannot continue with the reconcile loop because we error out on the operator image resolution. It's annoying when debugging locally.

With this change if we cannot find the operator deployment in the cluster, we return the default operator image


czunker (approved):
Thanks @imilchev

PR #1112, opened 2024-05-14
✨ allow specifying container proxy settings

imilchev:
expose a way of setting the configuration introduced in mondoohq/cnquery#3981 via the operator


controllers/mondoooperatorconfig_controller.go
```diff
@@ -59,8 +59,8 @@ func (r *MondooOperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{}, nil
 	}
 
-	if config.Spec.HttpProxy != nil {
-		urlParsed, err := url.Parse(*config.Spec.HttpProxy)
+	if config.Spec.ContainerProxy != nil {
```
preslavgerchev, line 62: is this change intended? we dont use http proxy here anymore?

imilchev: yup, it's intended. We don't want to mix up the http proxy setting with the container proxy one. We should route container traffic to a proxy only if `ContainerProxy` is set

preslavgerchev (approved):

PR #1137, opened 2024-06-24
🚀11.3.0

slntopp:


charts/mondoo-operator/templates/metrics-service.yaml-z
imilchev, line 1: I believe we never commit the -z files. @czunker can you confirm?

charts/mondoo-operator/templates/deployment.yaml
```diff
@@ -29,10 +29,12 @@ spec:
   replicas: {{ .Values.controllerManager.replicas }}
   selector:
     matchLabels:
+      app.kubernetes.io/name: mondoo-operator
     {{- include "mondoo-operator.selectorLabels" . | nindent 6 }}
   template:
     metadata:
       labels:
+        app.kubernetes.io/name: mondoo-operator
```
imilchev, line 37: Can we keep this one? I thought we had to remove it

imilchev: yup, I can confirm, these should be gone. I don't have them when I generate the release. Perhaps you missed to change some `sed` to `gsed`?

czunker: Here is the command which should remove it: https://github.com/mondoohq/mondoo-operator/blob/main/Makefile#L372

czunker (approved):
LGTM

Thanks @slntopp

PR #1141, opened 2024-06-25
🧹 fix repo name for GCP containers

imilchev:


.github/workflows/publish.yaml
```diff
@@ -16,7 +16,7 @@ on:
 env:
   REGISTRY: ghcr.io
   GHCR_IMAGE: ghcr.io/${{ github.repository }}
-  GCP_IMAGE:  us-docker.pkg.dev/mondoohq/release/${{ github.repository }}
```
imilchev, line 19: this one is equal to `mondoohq/mondoo-operator`, which is not what we want

czunker (approved):

PR #1153, opened 2024-07-30
✨ focus scanning on customer workloads

slntopp:


controllers/resource_monitor/resource_monitor_controller.go
```diff
@@ -82,8 +84,23 @@ func (r *ResourceMonitorController) Start(ctx context.Context) error {
 }
 
 func (r *ResourceMonitorController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	obj := r.createRes()
```
imilchev, line 87: this code looks correct to me. Please add some test cases in `controllers/resource_monitor/resource_monitor_controller_test.go`. Everything there should be already mocked, so it should be fairly simple to verify that we aren't add resources to the debouncer that have a parent we don't care about

slntopp: Updated the tests ✨

imilchev:
You will need to update the integration tests since they verify the workloads that end up in the mondoo platform. Since this now changed, the tests need to be updated to expect the correct workloads

slntopp:
@imilchev yep, working on it rn

imilchev (approved):
LGTM

PR #1284, opened 2025-11-14
Update Mondoo client image and tag in config

AdamVB:


imilchev:
this is not the right place to handle this check. I have updated the reference in #1305

PR #1303, opened 2025-12-03
Pin envtest / security workflow for golang 1.24

tas50:
Hold this until we're on 1.25

Integrations tests are still failing, but without this we don't even get to those tests as both unit tests and security tests fail.


czunker (approved):
Thanks @tas50

PR #1305, opened 2025-12-12
🧹 update cnquery/cnspec to v12

imilchev:
- update cnquery to v12
- update cnspec to v12
- update k8s provider to v12
- update `controller-runtime` to latest commit on `main`. We use unreleased version because the latest release does NOT support go v1.25.
- update all k8s modules
- update golangci-lint to latest version
- update default cnspec container image to v12


.golangci.yml
```diff
@@ -1,6 +1,18 @@
...
 # SPDX-License-Identifier: BUSL-1.1
-
+version: "2"
+run:
+  modules-download-mode: readonly
 linters:
+  default: none
   enable:
-    - gofumpt
+    #- errcheck
```
imilchev: i will enable the commented out linters in a follow-up. Those require some changes to the code to make the linters pass, that's why I want to split the changes

PR #1358, opened 2026-02-03
external cluster setup

chris-rock:
Updates documentation and adds an integration test for external cluster scanning.


Makefile
```diff
@@ -158,6 +158,9 @@ test/integration/ci: gotestsum load-minikube/ci
 endif
 	$(GOTESTSUM) --junitfile integration-tests.xml -- ./tests/integration/... -ldflags $(LDFLAGS) -v -timeout 20m -p 1
 
+test/integration/external-cluster: ## Run external cluster integration test (requires k3d management cluster)
```
imilchev: we probably want to include this in the github action that runs tests with k3d

PR #1379, opened 2026-02-06
⭐️ add annotations support for scanned assets

### LINKED ISSUE #955: add tags to scanned k8s objects
cnquery/cnspec allow tagging the scanned assets with user supplied values. But the operator currently does not offer this.

We should extend the `MondooAuditConfig`, so that a user can add tags, which then get added to every scanned asset.

Users could add labels to their k8s objects. But especially in larger clusters with different teams using it, this can be problematic. And I think it does not help with searching in the UI.

**Additional context**
-

chris-rock:
## Summary

- Adds `annotations` field to `MondooAuditConfigSpec` to allow users to define custom key-value pairs
- Annotations are attached to all scanned assets, making them searchable and filterable in the Mondoo Console
- Supports all scan types: K8s resources, nodes, container images, and resource watcher

## Usage Example

```yaml
kind: MondooAuditConfig
metadata:
  name: mondoo-client
spec:
  mondooCredsSecretRef:
    name: mondoo-credentials
  annotations:
    environment: production
    team: platform
    cost-center: "12345"
  kubernetesResources:
    enable: true
  nodes:
    enable: true
```

## Implementation

Annotations are passed to cnspec via:
- **Inventory files**: For K8s resources, nodes, and container image scans (using `Asset.AddAnnotations()`)
- **Command-line flags**: For the resource watcher (`--annotation key=value`)

## Test plan

- [x] `make lint` passes
- [x] `make test` passes
- [ ] Deploy operator with annotations configured and verify they appear on assets in Mondoo Console

Closes #955

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -29,9 +29,12 @@ func AnnotationArgs(annotations map[string]string) []string {
 	return args
 }
 
+const maxAnnotationLength = 256
+
 // Validate checks that annotation keys and values are well-formed for use as
-// cnspec --annotation key=value CLI arguments. Keys must be non-empty and must
-// not contain '='. Values must be non-empty.
+// cnspec --annotation key=value CLI arguments. Keys must be non-empty, must
+// not contain '=', and both keys and values must not exceed 256 characters.
+// Values must be non-empty.
 func Validate(annotations map[string]string) error {
 	for k, v := range annotations {
 		if k == "" {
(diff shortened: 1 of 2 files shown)


chris-rock:
## Future improvement: annotation validation should not skip full reconciliation

When annotations fail validation, the controller currently does `return ctrl.Result{}, nil` at the annotation check ([mondooauditconfig_controller.go:238](https://github.com/mondoohq/mondoo-operator/blob/add-annotations-to-scanned-assets/controllers/mondooauditconfig_controller.go#L238)), which skips ALL remaining reconciliation — not just annotation-related work. This means changes to schedules, enabling/disabling scan types, credential rotation, etc. are all blocked until annotations are fixed.

A future improvement would be to continue reconciliation with `Annotations: nil` (or the previous valid set) while in a degraded state, so that other spec changes are not blocked by an annotation typo.

PR #1381, opened 2026-02-06
⭐️ use Server-Side Apply (SSA) for resource reconciliation

### LINKED ISSUE #687: Use server-side-apply for the operator
-

We should give server side apply (SSA) a try for our operator.

From what I understood, this could reduce the code we have to maintain and make the reconcile more straight forward.

We can keep the operator as is. It works as it is.

**Additional context**
General docs about SSA:
https://kubernetes.io/docs/reference/using-api/server-side-apply/#using-server-side-apply-in-a-controller

Advantages to controllers:
https://kubernetes.io/blog/2022/10/20/advanced-server-side-apply/?utm_source=pocket_mylist#controller-authors

chris-rock:
## Summary

This PR replaces the previous custom Server-Side Apply (SSA) implementation with controller-runtime's upstream `CreateOrUpdate` pattern.

- Restore `pkg/utils/k8s/create_update.go` wrapper around `ctrl.CreateOrUpdate` with controller reference and logging
- Add `pkg/utils/k8s/update_fields.go` with field-level update helpers (`UpdateCronJobFields`, `UpdateDeploymentFields`, `UpdateDaemonSetFields`) that copy only managed fields, preserving server-set defaults
- Convert all handlers (container_image, k8s_scan, nodes, resource_watcher) to use `CreateOrUpdate` with field-level mutate functions
- Replace `DeleteAllOf` with `DeleteCompletedJobs` to avoid killing running scan Jobs
- Add `TerminationMessagePath`/`TerminationMessagePolicy` to all container specs to prevent unnecessary diffs against server defaults
- Delete custom `apply.go`/`apply_test.go` (SSA approach)

### Why the approach changed

The initial SSA implementation caused integration test failures because:
1. **Constant update loop** — `obj.Spec = desired.Spec` in mutate functions zeroed out server-set defaults (DNSPolicy, SchedulerName, TerminationGracePeriodSeconds, etc.), causing `DeepEqual` to find differences on every reconcile
2. **Aggressive job cleanup** — `DeleteAllOf` killed ALL Jobs (including running ones) on every detected "update", preventing scan pods from ever completing

The upstream `CreateOrUpdate` + field-level updates pattern avoids both issues by only mutating fields the operator manages, letting the API server retain its defaults.

Closes #687

## Test plan

- [x] All unit tests pass (`make test`)
- [x] Linting passes (`make lint`, `make lint/actions`)
- [ ] Integration tests pass on a real cluster
- [ ] Verify resources are created/updated correctly without constant update loops


imilchev (commented):
I think we should use ctrl.CreateOrUpdate. Here's an example https://github.com/kubernetes-sigs/controller-runtime/blob/2053ba3d414e7cf9b47ae407e5c860556b3113b0/examples/multiclustersync/main.go#L165

PR #1440, opened 2026-03-20
Fix spell check workflow permissions

tas50:
## Summary
- Add `actions: read` to the Report job so it can download artifacts from the spell check job
- Add `security-events: write` to the Report job for SARIF upload
- Add top-level `permissions` block for least-privilege defaults
- Align all job permissions with the reference pattern

## Test plan
- [ ] Verify spell check workflow runs and the Report job can post comments on a PR


slntopp:
Superseded by a different action

PR #1520, opened 2026-06-15
feat: suspend scheduled scans

MaxRink:
## Summary
- add suspend flags for scheduled Kubernetes resource, external cluster, container image, and node scan CronJobs
- wire the suspend settings into generated CronJobs while keeping resources present
- update CronJob reconciliation to manage spec.suspend changes
- document suspension for upgrade-hiatus workflows and add focused tests

## Tests
- make generate manifests
- go test ./controllers/k8s_scan ./controllers/container_image ./controllers/nodes ./pkg/utils/k8s -count=1
- go test ./api/v1alpha2 -count=1
- go test ./controllers/... -count=1
- go test ./api/v1alpha2 ./pkg/utils/k8s -count=1
- git -c core.fsmonitor=false diff --check

### DIFF
--- a/.github/actions/link-check/config.json
+++ b/.github/actions/link-check/config.json
@@ -1,3 +1,3 @@
 {
-  "aliveStatusCodes": [429, 200, 406]
+  "aliveStatusCodes": [200, 406, 429]
 }
--- a/api/v1alpha2/mondooauditconfig_types.go
+++ b/api/v1alpha2/mondooauditconfig_types.go
@@ -134,6 +134,9 @@ type KubernetesResources struct {
 	ContainerImageScanning bool `json:"containerImageScanning,omitempty"`
 	// Specify a custom crontab schedule for the Kubernetes resource scanning job. If not specified, the default schedule is used.
 	Schedule string `json:"schedule,omitempty"`
+	// Suspend pauses scheduled Kubernetes resource scan CronJobs without deleting the generated resources.
+	// External cluster scan CronJobs inherit this value and can also be paused individually.
+	Suspend bool `json:"suspend,omitempty"`
 
 	// ResourceWatcher configures real-time resource watching and scanning.
 	// When enabled, a deployment will be created that watches for K8s resource changes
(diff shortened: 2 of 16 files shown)


MaxRink:

MaxRink:
recheck

MaxRink:
Rebased onto latest `main` (tip `8953fc85`).

During the rebase, `main` had independently added `Suspend: ptr.To(m.Status.ScanningPaused)` to the scan CronJob specs (console pause/unpause support, #1551), which collided with this PR's user-configurable `Suspend` fields. Resolved by unioning both signals so a CronJob is suspended when *either* the console has paused scanning *or* the user set the config `suspend` flag:

- `controllers/k8s_scan/resources.go`: `Suspend: ptr.To(m.Spec.KubernetesResources.Suspend || m.Status.ScanningPaused)` and, for external clusters, `m.Spec.KubernetesResources.Suspend || cluster.Suspend || m.Status.ScanningPaused`
- `controllers/nodes/resources.go`: `m.Spec.Nodes.Suspend || m.Status.ScanningPaused`
- `controllers/container_image/resources.go`: `m.Spec.Containers.Suspend || m.Status.ScanningPaused`

`go build`, `go vet`, and the touched-package tests (`k8s_scan`, `nodes`, `container_image`, `pkg/utils/k8s`) all pass. `make manifests generate` produces no diff. All prior review comments (external-cluster suspend inheritance docs + OR-path test) were already addressed and the review bot approved.

slntopp (changes requested):
Nice feature. A few extra unit tests would be good to have:

1. **Unsuspend path**: Add a test per scan type where `Suspend` starts `true`, gets set back to `false` (with `ScanningPaused=false`), and assert `*cj.Spec.Suspend == false`. The OR logic makes this trivially correct, but an explicit test documents the contract and guards against future refactors.

2. **Interaction with `ScanningPaused`**: Add a test where `Suspend=false` but `ScanningPaused=true` — assert the CronJob is still suspended. This ensures the two mechanisms remain independent and neither accidentally shadows the other.

3. **Both flags clearing**: `Suspend=false` + `ScanningPaused=false` → `*Suspend == false`. Confirms the CronJob actually resumes when both pause sources are removed.

These are quick to add (same shape as the existing suspend tests, just flipping the bools) and would make the behavioral matrix fully explicit.

slntopp (approved):

PR #1530, opened 2026-06-19
fix: use inventory-file for node scanner

MaxRink:
## Summary
- replace node scanner Go-template inventory values with a stable node-name placeholder
- render the placeholder to a writable `/tmp` inventory file in a lightweight `render-node-inventory` init container
- run the main scanner container directly as `cnspec ... --inventory-file`, so custom scanner images only need `cnspec` and no longer need `/bin/sh` or `sed`
- resolve the render-helper image through the shared container image resolver so `imageRegistry`, `registryMirrors`, `imagePullSecrets`, and `skipContainerResolution` apply to the helper image too
- document the init-container behavior in the node-scanning user manual

## Review fixes
- removed the hard-coded upstream helper image from rendered workloads by passing a resolver-managed helper image into CronJob and DaemonSet builders
- added resolver and reconciliation regression tests for registry-mirrored helper images

## Validation
- `git diff --check`
- `go test ./controllers/nodes ./pkg/utils/mondoo ./pkg/utils/k8s ./controllers/status`
- `go build -o /tmp/mondoo-operator-node ./cmd/mondoo-operator/main.go`
- `make lint/actions`
- `make lint`
- `kubectl --context kind-mondoo-oidc-kind apply -k config/crd`
- `kubectl --context kind-mondoo-oidc-kind apply --server-side --dry-run=server -k config/rbac`
- `kubectl --context kind-mondoo-oidc-kind apply --server-side --dry-run=server -f config/samples/k8s_v1alpha2_mondooauditconfig.yaml`
- `kubectl --context kind-mondoo-oidc-kind apply --server-side --dry-run=server -f config/samples/k8s_v1alpha2_mondoooperatorconfig.yaml`
- targeted `kind-mondoo-oidc-kind` server-side dry-run for a node-scanning `MondooAuditConfig` plus registry-mirrored `MondooOperatorConfig`

## Notes
- The init helper uses the existing lightweight BusyBox helper image pattern. Pod-level `imagePullSecrets` still apply to both the helper and scanner containers.

- `go test ./pkg/utils/mondoo -run 'TestContainerImageResolverSuite/Test(ContainerImageUsesRegistryMirror|BusyBoxImageUsesRegistryMirror|ApplyImageRegistry)'`
- `go test ./controllers/nodes -run 'TestResources|TestCronJob_UsesInventoryFileFlag|TestDaemonSet_UsesInventoryFileFlag|TestDeploymentHandlerSuite/TestReconcile_(CreateCronJobs|CronJobUsesResolvedRenderImage|CreateDaemonSet)'`
- `go test ./controllers/nodes ./pkg/utils/mondoo ./pkg/utils/k8s ./controllers/status`
- `make lint/actions`
- `make lint`

### DIFF
--- a/controllers/nodes/deployment_handler.go
+++ b/controllers/nodes/deployment_handler.go
@@ -14,6 +14,7 @@ import (
 
 	"go.mondoo.com/mondoo-operator/api/v1alpha2"
 	"go.mondoo.com/mondoo-operator/pkg/client/mondooclient"
+	"go.mondoo.com/mondoo-operator/pkg/constants"
 	"go.mondoo.com/mondoo-operator/pkg/utils/k8s"
 	logutils "go.mondoo.com/mondoo-operator/pkg/utils/logger"
 	"go.mondoo.com/mondoo-operator/pkg/utils/mondoo"
@@ -91,6 +92,11 @@ func (n *DeploymentHandler) syncCronJob(ctx context.Context) error {
 		n.log().Error(err, "Failed to resolve mondoo-client container image")
 		return err
 	}
+	renderImage, err := n.ContainerImageResolver.ContainerImage(ctx, constants.BusyBoxImage, n.MondooOperatorConfig.Spec.SkipContainerResolution)
+	if err != nil {
+		logger.Error(err, "Failed to resolve node inventory render container image")
+		return err
+	}
 
 	clusterUid, err := k8s.GetClusterUID(ctx, n.KubeClient, n.log())
 	if err != nil {
@@ -125,7 +131,7 @@ func (n *DeploymentHandler) syncCronJob(ctx context.Context) error {
 			return err
 		}
 
-		desired := CronJob(mondooClientImage, node, n.Mondoo, n.IsOpenshift, *n.MondooOperatorConfig)
+		desired := CronJob(mondooClientImage, renderImage, node, n.Mondoo, n.IsOpenshift, *n.MondooOperatorConfig)
 		cronJob := &batchv1.CronJob{ObjectMeta: metav1.ObjectMeta{Name: desired.Name, Namespace: desired.Namespace}}
 		op, err := k8s.CreateOrUpdate(ctx, n.KubeClient, cronJob, n.Mondoo, n.log(), func() error {
 			k8s.UpdateCronJobFields(cronJob, desired)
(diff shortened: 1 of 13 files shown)


MaxRink:
Rebased onto `main` (8953fc85) — clean, no conflicts. All three prior review comments (busybox registry prefix in `pkg/constants/images.go`, test using `constants.BusyBoxImage`, and the `sed` delimiter note) were already addressed and are resolved; the final automated review is APPROVED.

Verification on the rebased branch:
- `go build ./...` ✅
- `go vet` + `go test ./controllers/nodes ./pkg/utils/mondoo ./pkg/utils/k8s ./controllers/status` ✅
- `make lint` (golangci-lint) — 0 issues ✅
- `make lint/actions` (actionlint) ✅
- `gofmt` clean, `git diff --check` clean

No API types changed, so no manifest regeneration needed. Force-pushed with lease (tip `d0929d25`).

slntopp (commented):
What's the motivation for moving from `--inventory-template` to an init container + `--inventory-file`? The current Go-template approach works, and adding an init container introduces a few concerns:

- **New runtime dependency**: Every node scan pod now needs to pull `busybox:1.36`. Air-gapped and registry-restricted environments need to mirror yet another image. We've had a fair share of issues with init containers / sidecars in the WIF setup already — this adds the same class of risk to the much more common node scanning path.
- **Failure visibility**: If the init container fails (image pull, OOM, sed error), pods sit in `Init:Error` with no operator-level degraded condition. Users have to manually inspect pods to find out why scanning stopped.
- **Blast radius**: This replaces the mechanism for ALL node scanning (CronJob and DaemonSet), with no fallback to the prior approach.

Is there a concrete problem with `--inventory-template` that this solves? The PR description mentions that custom scanner images no longer need `/bin/sh` or `sed` — are there users today running into issues with custom scanner images that lack a shell? If so, it would help to understand what those images look like and why they're built that way, so we can evaluate whether the init container tradeoff is the right fix versus, say, documenting the shell requirement for custom images.

MaxRink:
@slntopp I guess it mostly comes down to personal preference, we try to not have privileged pods around that have shells and other tooling available. 
With this the init container basically dies as soon as the scan starts.
In our case the mondoo scanners also run on nodes where I /really/ dont want someone to exfiltrate certain data from and having less tools available in the container thus was preferred.

PR #1556, opened 2026-07-06
fix: set priorityClassName

niklastreml:
# The bug
We deployed a mondooauditconfig with a priorityClass in our environment, but noticed the priorityClass was not set on any of the jobs spawned by the cronjobs. here is an auditconfig for reproduction:

 ```yaml
kind: MondooAuditConfig
metadata:
 name: audit
 namespace: mondoo-operator
spec:
 consoleIntegration:
   enable: true
 kubernetesResources:
   enable: false
 containers:
   enable: false
 nodes:
   enable: true
   priorityClassName: system-node-critical
   style: cronjob
 mondooCredsSecretRef:
   name: mondoo-client
 mondooTokenSecretRef:
   name: mondoo-token
 ```

I looked through the code a bit and noticed, that the priorityClass is never set when using the cronjob style. 

## Fix
I added a bare minimum fix in this pr. I guess this could also be refactored out into a function, so this sort of problem doesn't happen again. Let me know if that's something you want implemented and I'll do it.

### DIFF
--- a/controllers/nodes/resources.go
+++ b/controllers/nodes/resources.go
@@ -95,9 +95,10 @@ func CronJob(image string, node corev1.Node, m *v1alpha2.MondooAuditConfig, isOp
 							},
 						},
 						Spec: corev1.PodSpec{
-							NodeName:      node.Name,
-							RestartPolicy: corev1.RestartPolicyOnFailure,
-							Tolerations:   k8s.TaintsToTolerations(node.Spec.Taints),
+							NodeName:          node.Name,
+							PriorityClassName: m.Spec.Nodes.PriorityClassName,
+							RestartPolicy:     corev1.RestartPolicyOnFailure,
+							Tolerations:       k8s.TaintsToTolerations(node.Spec.Taints),
 							// The node scanning does not use the Kubernetes API at all, therefore the service account token
 							// should not be mounted at all.
 							AutomountServiceAccountToken: ptr.To(false),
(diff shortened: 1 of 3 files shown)


slntopp:
Thank you for your contribution, looks good to me, could you please update tests to include the `priorityClass`, e.g. 
- `controllers/nodes/resources_test.go` @ `TestCronJob_WithPriorityClassName`
- `pkg/utils/k8s/update_fields_test.go` @ `TestUpdateCronJobFields_ImagePullSecrets`, `TestUpdateCronJobFields_PreservesUnmanagedFields`

slntopp:
Thanks for reporting this! The bug was real at the time, but it's since been fixed on `main` and will be delivered with `13.4.0` — the CronJob path now sets `PriorityClassName` from `spec.nodes.priorityClassName` (see `controllers/nodes/resources.go:233`), which landed as part of the jobOverrides refactor in #1594.

On that note — we've also added `jobOverrides` as a more general mechanism for customizing scan workloads. If you need to set priority class (or tolerations, labels, annotations, TTL) across all scan types, you can now use `spec.jobOverrides` or per-section overrides like `spec.nodes.jobOverrides`. The dedicated `spec.nodes.priorityClassName` field still works, but `jobOverrides` is the more flexible path going forward.

This PR will be closed — feel free to verify on the latest version that it works for your setup.

Issue #1600, opened 2026-08-04
Node scan memory limit is not validated against node capacity, and NodeName pinning bypasses the scheduler that would catch it


MaxRink:
### Summary

Node-scan pods are pinned with `PodSpec.NodeName`, which bypasses the scheduler. That means the usual safety net — the scheduler refusing to place a pod whose requests do not fit — never runs, and a `spec.nodes.resources.limits.memory` larger than the node's allocatable memory is silently accepted and admitted by kubelet.

`controllers/nodes/resources.go`:

```go
Spec: corev1.PodSpec{
    NodeName: node.Name,
    ...
```

and the limit is passed straight through:

```go
containerResources := k8s.ResourcesRequirementsWithDefaults(m.Spec.Nodes.Resources, k8s.DefaultNodeScanningResources)
gcLimit := gomemlimit.CalculateGoMemLimit(containerResources)
```

`CalculateGoMemLimit` then sets `GOMEMLIMIT` to 90% of that limit, so `cnspec` is actively *instructed* to grow to nearly the configured size.

This is not a complaint about your defaults. `DefaultNodeScanningResources` is `250M` / `300m`, which is sensible. The problem is the failure mode when an operator overrides it with a value that does not fit the node — the operator has all the information needed to catch it, and nothing else in the stack will.

### Why the failure mode is unusually bad

If the configured limit is **at or above the node's memory capacity**, the pod's cgroup ceiling can never be reached before the node itself runs out of memory. The kernel therefore cannot contain the scanner:

* Below node capacity → the kernel performs a **cgroup-scoped** OOM kill. The scan dies, the node is fine:
  ```
  kernel: cnspec invoked oom-killer: ... oom_score_adj=970
  kernel: oom-kill:constraint=CONSTRAINT_MEMCG, ..., oom_memcg=/kubepods.slice/kubepods-burstable...
  kernel: Memory cgroup out of memory: Killed process (cnspec) total-vm:3170296kB, anon-rss:1835144kB
  ```
* At or above node capacity → no cgroup ceiling is ever hit, the kernel goes into **global** reclaim, and kubelet is starved. The node reports `Ready=Unknown` with `NodeStatusUnknown` / *"Kubelet stopped posting node status"* and does not recover on its own.

[...]

PR #1601, opened 2026-08-04
feat(nodes): warn when the node scan memory limit is too large for the node

MaxRink:
Draft, for discussion — happy to change the shape, the threshold, or drop it entirely if you'd rather solve this differently.

Addresses #1600.

### Why

Node scan pods are pinned with `PodSpec.NodeName`, so they bypass the scheduler and nothing validates `spec.nodes.resources.limits.memory` against the node the pod will actually run on. That matters more than it might sound, because of how the failure degrades:

* **Limit comfortably below node capacity** → an overrunning scan is killed by its own cgroup (`oom-kill:constraint=CONSTRAINT_MEMCG`). The scan fails loudly, the node is fine. This is the good case and it is self-announcing.
* **Limit at or above node capacity** → the pod's cgroup ceiling can never be reached before the *node* runs out of memory, so the kernel does global reclaim instead. kubelet gets starved, stops renewing its node lease, and the node goes `Ready=Unknown` / `NodeStatusUnknown` some time later. `GOMEMLIMIT` is derived from the same limit, so `cnspec` is actively instructed to grow into that space.

The second case is nasty precisely because it does *not* look like a scanner problem: there is no OOMKilled pod to notice, just a node that quietly stops posting status. We spent a day attributing it to a node image and a control-plane rollout before the cron boundary correlation pointed back at the scan.

The operator already lists `Node` objects to build these CronJobs, so `status.allocatable` is sitting right there at exactly the moment the decision is made — it seemed a shame not to use it.

### What this does

Adds `UnsafeNodeScanMemoryLimit()` plus a rendered warning, and logs it from the node CronJob reconcile loop when the configured limit exceeds **70%** of a node's allocatable memory. The message names the node, the configured limit and the node's allocatable memory, so it is actionable without going back to the cluster:

```
Node scan memory limit (8G) is 103% of the node's allocatable memory (8028136Ki).
Node scan pods are pinned with spec.nodeName and bypass the scheduler, so this limit is not
validated against the node. At this size the kernel may be unable to contain an overrunning
scan within its own cgroup and can starve kubelet instead of OOMKilling the scan.
Consider lowering spec.nodes.resources.limits.memory to at most 70% of the smallest node's
allocatable memory.
```

Deliberately minimal:

* **No API change** and **no behaviour change** — advisory only.
* `DefaultNodeScanningResources` (250M) never trips it, so out-of-the-box users see nothing.
* Missing/zero limit and missing/zero `allocatable` are both treated as "nothing to say" rather than as errors.
* 70% is intentionally permissive: it is meant to catch *"this limit is the size of the node"*, not to second-guess reasonable tuning.

### Tests

`controllers/nodes/resource_validation_test.go`, table-driven, covering the operator default on a small node, limit == node, limit > node, the same limit on a large node, either side of the threshold, and both "nothing to compare" cases. Also asserts the rendered message contains both quantities and the field name.

```
$ go test ./controllers/nodes/...
ok  	go.mondoo.com/mondoo-operator/controllers/nodes
```

`go build ./...` is clean and `gofmt` reports nothing.

### Open questions for you

1. **Log spam.** This currently logs once per node per reconcile. That is fine at `Info` for a condition that should be rare, but if you would prefer it deduplicated, hoisted out of the per-node loop, or emitted once per `MondooAuditConfig` rather than per node, say the word.
2. **Event or condition instead?** #1600 lists emitting a Kubernetes `Event` or a `MondooAuditConfig` status condition as alternatives. Both are more discoverable than a log line, but they touch your status API surface, so I did not want to choose that for you. Happy to switch.
3. **Threshold.** 70% is a judgement call. Anything from 50% to 80% would be defensible.
4. **Clamping.** The stronger version — capping the effective limit (and therefore `GOMEMLIMIT`) at a fraction of allocatable so the bad configuration becomes impossible rather than merely visible — would need to be opt-in. Out of scope here, but I am glad to follow up if you want it.

### Note

I have not signed the Mondoo CLA yet; I will sign it via the CLA Assistant bot when it comes round. Also worth flagging that our own misconfigured 8G limit was **our** mistake, not a bad default on your side — `DefaultNodeScanningResources` is conservative and sensible. This is about making that class of mistake visible, not about changing your defaults.

### DIFF
--- a/controllers/nodes/deployment_handler.go
+++ b/controllers/nodes/deployment_handler.go
@@ -8,6 +8,7 @@ import (
 	"maps"
 	"slices"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -20,6 +21,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -113,6 +115,15 @@ func (n *DeploymentHandler) syncCronJob(ctx context.Context) error {
 		return err
 	}
 
+	// Collected across all nodes so the warning is emitted once per reconcile
+	// instead of once per affected node.
+	var (
+		unsafeMemoryNodes       []string
+		unsafeMemoryLimit       resource.Quantity
+		unsafeMemoryAllocatable resource.Quantity
+		haveUnsafeMemory        bool
+	)
+
 	// Create/update CronJobs for nodes
 	for _, node := range nodes.Items {
 		cm := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: ConfigMapNameWithNode(n.Mondoo.Name, node.Name), Namespace: n.Mondoo.Namespace}}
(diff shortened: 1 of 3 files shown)


MaxRink:

PR #1617, opened 2026-08-11
fix(node-scanning): sort aggregated tolerations for deterministic DaemonSet diffs

msprung91:
## What
`syncDaemonSet` builds the node-scanning DaemonSet's tolerations from the union of
every distinct taint across all cluster nodes, deduplicated via a `map` and collected
with `slices.Collect(maps.Keys(...))`.

## Bug
Go randomizes map iteration order, so the resulting slice order differs across
reconciles even when the underlying taint set hasn't changed. Since
`equality.Semantic.DeepEqual` compares slices order-sensitively, `CreateOrUpdate` saw
a spurious diff on (almost) every reconcile, triggering a DaemonSet update and a full
rolling restart of every node-scanning pod, cluster-wide, on every reconcile. This gets
worse the more distinct taint keys exist in the cluster (observed via a DaemonSet
`.metadata.generation` in the tens of thousands after just a few days).

## Fix
Sort the collected tolerations (`SortTolerations`, `pkg/utils/k8s/taints.go`) before
building the desired DaemonSet, so an unchanged taint set always produces an
identical, order-independent slice.

## Testing
- `TestSortTolerations_Deterministic` feeds 4 permutations of the same toleration set
  and asserts they all converge on the same sorted output.
- `go build`, `go vet`, `go test ./pkg/utils/k8s/... ./controllers/nodes/...` all pass.

### DIFF
--- a/controllers/nodes/deployment_handler.go
+++ b/controllers/nodes/deployment_handler.go
@@ -246,7 +246,11 @@ func (n *DeploymentHandler) syncDaemonSet(ctx context.Context) error {
 		}
 	}
 
-	desired := DaemonSet(*n.Mondoo, n.IsOpenshift, mondooClientImage, *n.MondooOperatorConfig, slices.Collect(maps.Keys(tolerations)))
+	// maps.Keys iterates in random order, so sort to avoid a spurious update on every reconcile.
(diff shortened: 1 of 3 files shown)


msprung91:

msprung91:
recheck

PR #1632, opened 2026-08-26
🐛 nodes: give node assets their cloud instance identity

### LINKED ISSUE #1630: Node scan pins a k8s-only platform ID, so node assets never dedup with their cloud VM asset
## Summary

The node scanning inventory pins `PlatformId` to a k8s-scoped identifier on a `filesystem` connection. Because the OS provider treats an explicitly supplied filesystem platform ID as the asset's *complete* identity, the hostname and cloud-detect ID detectors never run for node scans.

Consequence: a node that is also discovered by a cloud integration (AWS EC2, GCE, Azure VM) ends up as **two separate assets for the same machine** — scanned twice, scored twice, findings counted twice — with no relationship linking them. The platform's asset dedup is keyed purely on platform identifiers, and the two ID sets are disjoint, so dedup can never fire.

## Where

`controllers/nodes/resources.go`, `Inventory()`:

```go
Connections: []*inventory.Config{
    {
        Type:       "filesystem",
        Host:       "/mnt/host",
        PlatformId: fmt.Sprintf(`{{ printf "//platformid.api.mondoo.app/runtime/k8s/uid/%%s/node/%%s" "%s" (getenv "NODE_NAME")}}`, clusterUID),
    },
},
```

In [mondoohq/mql](https://github.com/mondoohq/mql), `providers/os/provider/provider.go`, the `Type_FileSystem` branch of `Service.connect`:

```go
case shared.Type_FileSystem.String():
    conn, err = fs.NewConnection(connId, conf, asset)
    ...
    pID, err := conn.(*fs.FileSystemConnection).Identifier()
    if err != nil {
        fingerprint, p, err := id.IdentifyPlatform(conn, req, asset.Platform, asset.IdDetector)
        ...
    } else {
        // In this case asset.Name should already be set via the inventory
        asset.PlatformIds = []string{pID}
    }
```

`fs.FileSystemConnection.Identifier()` (`providers/os/connection/fs/filesystem.go`) returns exactly the injected ID and errors only when none was supplied — so for operator node scans the `else` branch always wins. That skips `id.IdentifyPlatform`, which is what runs the default detectors (`providers/os/provider/detector.go`, `IdDetectors` = `Hostname` + `CloudDetect`). Note also that `asset.PlatformIds` is **replaced**, not appended to.

Because `CloudDetect` never runs, the related-asset links it would normally emit (`clouddetect` → `RelatedPlatformIDs`) are lost as well, so the two assets aren't even linked to each other.

## Why this is a dedup bug, not a missing merge

Worth stating explicitly, because the natural reading of "two assets for one machine" is "the platform should merge them" — it won't, and it has no facility to.

There is **no merge operation anywhere in the platform**: no merge API, no GraphQL mutation, no background job that reconciles duplicate assets after the fact. Asset identity is resolved once, at ingest, in `findAssetsByPlatformIdentifiers` (`policy/datalakes/db/assetstore.go`): the incoming scan's platform IDs are looked up against `platform_identifiers`, and the scan lands on whichever existing asset already owns one of them. A shared ID means one asset; no shared ID means a new asset, permanently.

Two consequences that shape the fix:

1. **The identifier has to be emitted at scan time.** Nothing downstream can infer that a k8s node ID and an EC2 instance ID describe the same box.
2. **Existing duplicates are not repaired retroactively.** Once the node scan starts emitting the cloud instance ID, subsequent scans dedup onto the cloud VM asset — but the already-created duplicate rows stay until they age out.

The one code path that *is* called a merge — `reconcileThirdPartyIntoCnspecAsset` (`policy/datalakes/db/sbom.go`) — does not apply here. It is SBOM-import-only, skips cnspec-generated SBOMs outright, and its `shouldAttemptHostnameMerge` guard requires at least one `//platformid.api.mondoo.app/hostname/` identifier. A k8s node ID is neither a hostname ID nor one of the known third-party correlation prefixes, so the guard returns false immediately.

The dedup priority groups (same file) already do the right thing once both IDs are present: a cloud instance ID sits in the strongest priority tier, above hostname, so a node reporting both IDs attaches to the existing cloud VM asset rather than creating a third one.

## Observed

One Bottlerocket EKS node, scanned both by the operator's node scan and by a cloud integration. Two assets (identifiers redacted):

```
asset A   name: ip-10-x-x-x.<region>.compute.internal        type: machine
          //platformid.api.mondoo.app/runtime/k8s/uid/<cluster-uid>
            /node/ip-10-x-x-x.<region>.compute.internal
          labels: k8s.mondoo.com/kind=node, distro-id=bottlerocket
          relatedAssets: 0

asset B   name: <ec2 Name tag>                               type: aws.ec2_instance
          //platformid.api.mondoo.app/runtime/aws/ec2/v1/accounts/<account-id>
            /regions/<region>/instances/<instance-id>
          relatedAssets: 0
```

Both are cnspec-scanned, both run the same policies, both are scored independently, and both report zero related assets.

This is not a one-off. In one production space every Bottlerocket EKS worker node is duplicated: **345 `aws.ec2_instance` assets alongside 360 `machine` assets**, same OS build on both sides, ~705 assets for ~350 machines. Pairs were confirmed by joining the EC2 asset's reported private IP to the k8s node name (15 exact pairs in a 40-asset sample; the remainder report a `100.64.x` pod-CIDR secondary address, so the join simply can't be made for those — not evidence they are unpaired).

Nothing here is EC2-specific — the same shape applies to GCE and Azure VMs, and to any node the operator scans that a cloud integration also discovers. It scales with cluster size: every node in every such cluster is duplicated.

## Expected

A node asset should carry both its k8s-scoped identifier *and* the cloud instance identifier (or at minimum emit the cloud instance ID as a related asset), so that a node discovered by both the operator and a cloud integration dedups to a single asset.

## Possible directions

- **Append instead of replace** in the filesystem branch: keep the inventory-supplied ID *and* run the cloud-detect detector, so node assets report both IDs. Worth noting the filesystem connection has no command execution (`RunCommand` is not implemented), so detection would have to work from the mounted host filesystem and/or IMDS over HTTP from the scan pod.
- **Read it off the node object.** The operator already talks to the API server, and `node.spec.providerID` carries the instance identity directly (`aws:///<az>/i-…`, `gce://<project>/<zone>/<name>`, Azure's resource ID). Templating a second platform ID from that avoids the detector problem entirely — no IMDS, no command execution.
- Alternatively, expose an operator-level option to enable additional ID detectors for node scans, rather than changing the default.
- One thing to verify if this is implemented: the identifier set is replaced on each sync, so a cloud-integration scan reporting only the instance ID would drop the k8s ID again, and the next node scan would re-add it — worth confirming that flip-flop is harmless.

## Workaround

Exclude Kubernetes nodes from cloud-integration scanning (e.g. filter on the Karpenter / `kubernetes.io/cluster/<name>` tags), leaving the operator as the single source of truth for node assets. This avoids the duplication but gives up the cloud integration's view of those instances.

AdamVB:
Fixes #1630

> [!IMPORTANT]
> Depends on **mondoohq/mql#10484**. Merging this alone does not fix the issue — the identifier this PR produces is discarded until cnspec ships the matching provider change.

## The problem

A node the operator scans and a cloud integration also discovers ends up as two assets for one machine. Asset identity is keyed on platform identifiers, and the two sides share none: the node scan reports a Kubernetes identifier, the cloud integration an EC2 one. Nothing downstream can tell they describe the same host, and nothing reconciles duplicates after the fact — identity is decided once, when the scan is ingested.

Observed in one production space: **345 `aws.ec2_instance` assets alongside 360 `machine` assets** for ~350 EKS worker nodes, same OS build on both sides. Confirmed pairs by joining the EC2 asset's reported private IP to the k8s node name.

## Why the node scan cannot discover this itself

It connects over a filesystem connection to a bind-mounted host root, which has no command execution and — on a distro without cloud-init, such as Bottlerocket — no on-disk markers either.

The instance metadata service would answer, but a scan pod that is not on the host network cannot reach it. **Karpenter and the EKS managed node group module both default `httpPutResponseHopLimit` to 1.** Raising it hands the node role to every pod on the node, and host networking is a heavy grant for a scan pod.

## What this does instead

None of that is necessary. Two thirds of the identity are already in the Kubernetes API:

| piece | source | needs IMDS? |
|---|---|---|
| instance ID | `node.spec.providerID` → `aws:///eu-central-1a/i-0a8caeccfd813a595` | no |
| region | `topology.kubernetes.io/region`, else derived from the zone | no |
| account ID | resolved by the operator (below) | no |

Only the account is missing, and the operator can establish it without IMDS — first match wins:

1. **`AWS_ROLE_ARN`**, which IRSA injects into the operator pod. Free, no call.
2. **`sts:GetCallerIdentity`** — authoritative, and the only source that works under EKS Pod Identity. It needs no IAM permission (any valid credentials may call it), so this adds no privilege requirement, only the need for credentials to exist.
3. **An IRSA annotation on a ServiceAccount in the operator's namespace** — needs no AWS credentials at all, so it covers deployments where only the scanning ServiceAccount was ever wired up to IAM.

Resolved once per namespace and cached, including a negative answer, so a cluster outside AWS does not retry on every reconcile.

The identifier travels to the scan pod as `MONDOO_AWS_PLATFORM_ID` and reaches the provider through `inject-platform-ids`, which **appends rather than replaces** — the node keeps its Kubernetes identity and gains its cloud one.

## Everything fails closed

An unknown account, a non-EC2 provider, a Fargate node, or an availability zone whose region cannot be derived all produce no identifier and no environment variable, leaving the node scanned exactly as it is today.

That is deliberate. A wrong account or region does not fail loudly — it silently merges a node with a *different* machine's asset. Returning nothing is always the better answer.

## Scope

- **cronjob-style node scanning only.** The daemonset style builds one pod spec for every node, so a per-node value cannot be injected there.
- `go.mod` only flips `aws-sdk-go-v2/config` and `service/sts` from indirect to direct. No new modules.
- No new RBAC: the operator already reads nodes and ServiceAccounts.

## Tests

`controllers/nodes/aws_platform_id_test.go`, `controllers/nodes/aws_account_test.go` — Fargate task IDs, local zones (`us-west-2-lax-1a`, where trimming the trailing letter would invent a region that does not exist), GCP providerIDs, two- and three-slash providerID spellings, govcloud/china role ARNs, cross-namespace ServiceAccounts, and cache behaviour.

## Verifying

On a node with the matching cnspec build, the asset should carry both identifiers and resolve to the same asset as the cloud integration:

```
//platformid.api.mondoo.app/runtime/k8s/uid/<cluster-uid>/node/<node-name>
//platformid.api.mondoo.app/runtime/aws/ec2/v1/accounts/<account>/regions/<region>/instances/<instance-id>
```

### DIFF
--- a/controllers/nodes/aws_account.go
+++ b/controllers/nodes/aws_account.go
@@ -1,0 +1,174 @@
+/*
+Copyright 2022 Mondoo, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nodes
+
+import (
+	"context"
+	"os"
+	"regexp"
+	"sync"
+	"time"
+
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// irsaRoleAnnotation is the annotation IRSA uses to bind a ServiceAccount to an
+// IAM role.
+const irsaRoleAnnotation = "eks.amazonaws.com/role-arn"
+
+// stsTimeout bounds the GetCallerIdentity call. The result is cached for the
+// life of the process, so this runs at most once; a cluster with no route to
+// STS should not stall a reconcile waiting for it.
+const stsTimeout = 5 * time.Second
+
+// iamRoleARNRegex extracts the account ID from an IAM role ARN, e.g.
+// arn:aws:iam::123456789012:role/some-role. It accepts every partition, so
+// aws-cn and aws-us-gov ARNs work too.
+var iamRoleARNRegex = regexp.MustCompile(`^arn:aws[a-z-]*:iam::(\d{12}):`)
+
+// awsAccountResolver finds the AWS account the cluster runs in, without going
+// through the instance metadata service.
+//
+// IMDS is the obvious source and the wrong one here: Karpenter and the EKS
+// managed node group module both default httpPutResponseHopLimit to 1, which
+// stops any pod that is not on the host network from reading it. Raising that
+// limit exposes the node role to every pod on the node, and host networking is
+// a heavy grant for a scan pod. Every source below avoids the problem instead.
+//
+// The account never changes for a given operator deployment, so the answer is
+// resolved once and cached -- including a negative answer, which stops a
+// cluster with no AWS identity from retrying on every reconcile.
+type awsAccountResolver struct {
+	// Namespace is the operator's own namespace, where the ServiceAccount
+	// fallback looks for IRSA annotations.
+	Namespace string
+
+	once    sync.Once
+	account string
+}
+
+// awsAccountResolvers caches one resolver per operator namespace. The account
+// is a property of the cluster, so the answer is worth keeping across
+// reconciles -- the DeploymentHandler that asks for it is rebuilt on every one.
+// Keyed by namespace rather than held as a single global because the fallback
+// branch looks for ServiceAccounts in a specific namespace, and an operator may
+// reconcile MondooAuditConfigs in more than one.
+var awsAccountResolvers sync.Map
+
+// awsAccountResolverFor returns the cached resolver for an operator namespace.
+func awsAccountResolverFor(namespace string) *awsAccountResolver {
+	if resolver, ok := awsAccountResolvers.Load(namespace); ok {
+		return resolver.(*awsAccountResolver)
+	}
+	resolver, _ := awsAccountResolvers.LoadOrStore(namespace, &awsAccountResolver{Namespace: namespace})
+	return resolver.(*awsAccountResolver)
+}
+
+// Resolve returns the AWS account ID, or "" when it cannot be established. An
+// empty result is a normal outcome -- a cluster outside AWS, or one whose
+// operator has no cloud identity -- and callers must treat it as "skip the AWS
+// platform ID", never as an error worth failing a reconcile over.
+func (r *awsAccountResolver) Resolve(ctx context.Context, c client.Client) string {
+	r.once.Do(func() {
+		r.account = r.resolve(ctx, c)
+	})
+	return r.account
+}
+
+func (r *awsAccountResolver) resolve(ctx context.Context, c client.Client) string {
+	// 1. IRSA injects AWS_ROLE_ARN into every pod it serves, so the account is
+	//    already in this process's environment. Free, and no network call.
+	if account := accountFromRoleARN(os.Getenv("AWS_ROLE_ARN")); account != "" {
+		return account
+	}
+
+	// 2. Ask STS. This is the authoritative answer and the only one that works
+	//    under EKS Pod Identity, which hands out credentials without ever
+	//    naming a role ARN in the environment. GetCallerIdentity needs no IAM
+	//    permission at all -- any valid credentials can call it -- so this adds
+	//    no privilege requirement, only the need for credentials to exist.
+	if account := accountFromSTS(ctx); account != "" {
+		return account
+	}
+
+	// 3. Fall back to reading an IRSA annotation off a ServiceAccount in the
+	//    operator's namespace. This needs no AWS credentials at all, which
+	//    makes it the branch that works on a deployment where only the
+	//    scanning ServiceAccount was ever wired up to IAM.
+	//
+	//    It is last because it is the weakest: it reports the account of
+	//    whatever role somebody annotated, which is the cluster's account in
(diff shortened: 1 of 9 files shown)


AdamVB:
Tightened the ARN partition regex (used a non-capturing group so the account stays capture group 1). Declining the singleflight suggestion: disabling the SDK's IMDS probe made the STS call fast (test went 5.00s to 0.00s), reconciles are serialized per object, and a waiter gets the same answer singleflight would give it — not worth the extra dependency for a once-per-process lookup.

AdamVB:
**Server-side impact analysis** of this PR paired with mondoohq/mql#10484: https://claude.ai/code/artifact/4f3bca7b-7e76-4e28-857d-d35f83d5973b

TL;DR from a code-level trace of server/cnspec/operator/serverless-scanner-aws:

- **The identifier does what it promises**: the server resolves the dual-ID node scan onto the EC2 asset the AWS integration already manages (platform-ID priority groups, EC2 = 3 > k8s = 2), and node reporting keeps working.
- **Two things to plan for at rollout**: (1) the server *truncates* the stored ID set to the top priority group, so the k8s node ID is never persisted and the pre-existing `machine` asset is orphaned silently — ~350 assets going stale at once in the space from #1630; (2) the merged asset now has two full scan producers, and several server paths assume one writer — the resolved-policy sweep, the shared VEX/SBOM source, and the operator's own `PurgeAssets` GC filter (`managed_by` + `k8s.mondoo.com/kind: node` + 2×-interval staleness), which can purge the shared asset when a node leaves the cluster while the instance lives.
- **Merging is safe in any order** — both PRs fail closed alone. Recommendation: gate the operator release rollout on the server hardening listed in the page, and pair it with a one-time cleanup of orphaned k8s-keyed node assets.

